PR #2, opened 2017-09-29
Unstable

aardschok:
* Added collector and validator for renderer per layers
* Removed fix for `googleclientapi`


colorbleed/maya/lib.py
```diff
@@ -275,6 +275,11 @@ def collect_animation_data():
     return data
 
 
+def get_renderer(renderlayer):
+    cmds.editRenderLayerGlobals(currenRenderLayer=renderlayer)
```
BigRoy: This should be a context manager so that it reverts afterwards, or preferrably even get the value without changing the layer... for which there was a hack in older Maya's but that one doesn't work with the newer render setup systems.

PR #11, opened 2017-10-05
Fix ValidateRenderSettings and implement ValidateRenderImageRule

BigRoy:
> Note: This is **not** tested in production

### DIFF
--- a/colorbleed/plugins/maya/publish/validate_rendersettings.py
+++ b/colorbleed/plugins/maya/publish/validate_rendersettings.py
@@ -39,9 +39,9 @@ class ValidateResources(pyblish.api.InstancePlugin):
             padding = cmds.getAttr("defaultRenderGlobals.fileNamePadding")
 
         filename_prefix = default_prefix[renderer]
-        assert padding == default_padding, AttributeError(
+        assert padding == default_padding, (
             "Expecting padding of 4 ( 0000 )"
         )
-        assert prefix == filename_prefix, AttributeError(
+        assert prefix == filename_prefix, (
             "Wrong file name prefix, expecting %s" % filename_prefix
         )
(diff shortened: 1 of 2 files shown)


colorbleed/plugins/maya/publish/validate_rendersettings.py
```diff
@@ -21,27 +22,31 @@ class ValidateResources(pyblish.api.InstancePlugin):
...
+        # Collect the filename prefix in the renderlayer
+        with lib.renderlayer(layer_node):
+            if renderer == "vray":
+                prefix = cmds.getAttr("vraySettings.fileNamePrefix")
+                padding = cmds.getAttr("vraySettings.fileNamePadding")
+            else:
+                prefix = cmds.getAttr("defaultRenderGlobals.fileNamePrefix")
+                padding = cmds.getAttr("defaultRenderGlobals.fileNamePadding")
+
+        filename_prefix = self.DEFAULT_PREFIX[renderer]
```
aardschok: I think we should use `.get(renderer, "<Scene>/<Scene>_<RenderLayer>/<RenderLayer>")` to ensure it will pass when submitting with Maya Software for example.

aardschok: dict.get(renderer, "<Scene>/<Scene>_<RenderLayer>/<RenderLayer>") would be nice, we would support Maya native renderers for submission.

BigRoy: Not sure about this comment, but is this a double from the previous comment? :)

BigRoy: Yes, we should have a default. 

Actually, Arnold behaves just like the default renderers; only V-Ray seems to be behave differently. Maybe we should just default to `default` always, unless there's specific behavior to be implemented, like for V-Ray. That way the other behavior isn't linked to "arnold", but just used as default.

PR #15, opened 2017-10-16
Lkd 0006

aardschok:
* Improved documentation
* Remove function
* Moved logic to for-loop
* Removed unused `listHistory` command

### DIFF
--- a/colorbleed/plugins/maya/publish/collect_look.py
+++ b/colorbleed/plugins/maya/publish/collect_look.py
@@ -30,14 +30,16 @@ def get_look_attrs(node):
     result = cmds.listAttr(node, userDefined=True,
                            changedSinceFileOpen=True) or []
 
+    # `cbId` is added when a scene is saved, ignore by default
+    if "cbId" in result:
+        result.remove("cbId")
+
     # For shapes allow render stat changes
     if cmds.objectType(node, isAType="shape"):
         attrs = cmds.listAttr(node, changedSinceFileOpen=True) or []
-        valid = [attr for attr in attrs if attr in SHAPE_ATTRS]
-        result.extend(valid)
-
-    if "cbId" in result:
-        result.remove("cbId")
+        for attr in attrs:
+            if attr in SHAPE_ATTRS:
+                result.append(attr)
 
     return result
 
@@ -80,7 +82,7 @@ class CollectLook(pyblish.api.InstancePlugin):
         self.log.info("Gathering sets..")
         sets = self.collect_sets(instance)
 
-        # Lookup with absolute names (from root namespace)
+        # Lookup set (optimization)
         instance_lookup = set(cmds.ls(instance, long=True))
 
         self.log.info("Gathering set relations..")
@@ -88,12 +90,10 @@ class CollectLook(pyblish.api.InstancePlugin):
         # dict within the loop
         for objset in list(sets):
             self.log.debug("From %s.." % objset)
-            objset_members = sets[objset]["members"]
 
             # Get all nodes of the current objectSet
             for member in cmds.ls(cmds.sets(objset, query=True), long=True):
                 member_data = self.collect_member_data(member,
-                                                       objset_members,
                                                        instance_lookup)
                 if not member_data:
                     continue
@@ -109,7 +109,6 @@ class CollectLook(pyblish.api.InstancePlugin):
                 sets.pop(objset, None)
 
         self.log.info("Gathering attribute changes to instance members..")
-
         attributes = self.collect_attributes_changed(instance)
 
         # Store data on the instance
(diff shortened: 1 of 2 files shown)


colorbleed/plugins/maya/publish/collect_look.py
```diff
@@ -89,20 +86,25 @@ def collect(self, instance):
...
                     continue
 
+                # Add information of the node to the members list
                 sets[objset]["members"].append(member_data)
 
-        # Remove sets that didn't have any members assigned in the end
-        sets = self.remove_sets_without_members(sets)
+            # Remove sets that didn't have any members assigned in the end
+            # Thus the data will be limited to only what we need.
+            if not sets[objset]["members"]:
```
BigRoy: Great! Much easier to my eyes without needing to go to a separate function. :)

The only issue here is you can't pop the key from the dictionary as you are currently iterating over it. Otherwise it would behave similar to a list iterator where it then suddenly skips an index. In Python 2.7 this will even raise an error on dictionaries:

```python
a = {"a": 1, "b": 2, "c": 3}
for key in a:
    a.pop(key)
# # RuntimeError: dictionary changed size during iteration # 
```

A solution would to iterate over `a.keys()` (which makes a copy of the keys) instead of the `dict` itself. :) Though, I'm not sure how that works in Python 3+ because `dict.keys()` there I believe also returns an iterator. Feel free to test there too and find a behavior that works.

colorbleed/plugins/maya/publish/collect_look.py
```diff
@@ -118,6 +120,7 @@ def collect(self, instance):
         # Collect file nodes used by shading engines (if we have any)
         files = list()
         if looksets:
```
BigRoy: [Line 112](https://github.com/Colorbleed/colorbleed-config/pull/15/files#diff-3813032214deb50f51eaf9210aebdd55R112) gets the looksets by long names, but I believe `objectSets` are always their full path, because they cannot be i n a dag hierarchy. ;) So using `cmds.ls(x, long=True)` over the sets should be redundant...

Looking over the code here the order also looks slightly inlogical. Should we shift `looksets = cmds.ls(sets.keys(), long=True)` down and the logging of the found sets more up; so they are closer to their logic.

```python
self.log.info("Found the following sets:\n{}".format(sets.keys()))
```
This could be right below the collection of the sets (above attribute collection).

And the:
```python
looksets = cmds.ls(sets.keys(), long=True)
# which could also be:
looksets = sets.keys()
```
Could be directly above the `if looksets:` line.

colorbleed/plugins/maya/publish/validate_look_sets.py
```diff
@@ -19,6 +19,16 @@ class ValidateLookSets(pyblish.api.InstancePlugin):
...
 
+    Displacement shaders:
+    Ensure all geometry is added to the displacement objectSet.
+    It is best practice to add the transform group of the shape to the
+    displacement objectSet
+    Example content:
+        [asset_GRP|geometry_GRP|body_GES,
+         asset_GRP|geometry_GRP|L_eye_GES,
+         asset_GRP|geometry_GRP|R_eye_GES,
+         asset_GRP|geometry_GRP|wings_GEOShape]
```
BigRoy: This example seems off, you're recommending:

> It is best practice to add the transform group of the shape to the displacement objectSet

But `asset_GRP|geometry_GRP|wings_GEOShape` is not a transform, but the shape. So the recommended example should contain: `asset_GRP|geometry_GRP|wings_GEO`

colorbleed/plugins/maya/publish/collect_look.py
```diff
@@ -205,13 +187,7 @@ def collect_member_data(self, member, objset_members, instance_members,
             self.log.error("Node '{}' has no attribute 'cbId'".format(node))
             return
 
-        member_data = {"name": node, "uuid": lib.get_id(node)}
-
-        # Include components information when components are assigned
-        if components:
```
BigRoy: I don't see a need to remove the `components` from collection; they are part of the sets data. Even if we'd recommend avoiding things like face assignments I don't see a reason to break this particular logic at this point. Unless there's a strong argument to support this removal let's keep it in for now and preserve this logic.

colorbleed/plugins/maya/publish/collect_look.py
BigRoy: I've tested this, even with `dict.keys()` it will not work in Python 3 because it's an iterator, just like iterating over the dictionary itself.

The solution is to explicitly turn it into a copy of the keys using `list`, e.g.
```python
a = {"a": 1, "b": 2, "c": 3}
for key in list(a):
    a.pop(key)
```
This will work in both python 2 and 3.

colorbleed/plugins/maya/publish/collect_look.py
```diff
@@ -206,12 +190,10 @@ def collect_member_data(self, member, objset_members, instance_members,
             return
 
         member_data = {"name": node, "uuid": lib.get_id(node)}
-
-        # Include components information when components are assigned
         if components:
             member_data["components"] = components
 
-        return member_data
+        return member
```
BigRoy: Shouldn't this still return the member data? This will break things.

colorbleed/plugins/maya/publish/validate_look_sets.py
BigRoy: Note that issue is still open.

PR #17, opened 2017-10-18
Fixed REN-0002

aardschok:
* Remove filename from preview filename, 
* Added separator replace


BigRoy:
This is also implementing a `get_container_transforms` and changing some other code in that area. At first the actual logic doesn't seem to be altered so I'm assuming this PR should be fine. 

Feel free to merge if you believe this is good.

PR #18, opened 2017-10-19
Add safety logic for rendering

aardschok:
* Force save plugin before submitting rendering
* Incremental save scene after submitting render
* Separate validate Deadline plugin

### DIFF
--- a/colorbleed/plugins/maya/publish/increment_current_file_deadline.py
+++ b/colorbleed/plugins/maya/publish/increment_current_file_deadline.py
@@ -9,8 +9,8 @@ class IncrementCurrentFileDeadline(pyblish.api.ContextPlugin):
 
     """
 
-    label = "Increment current file.."
-    order = pyblish.api.IntegratorOrder + 99.0
+    label = "Increment current file"
+    order = pyblish.api.IntegratorOrder + 9.0
     hosts = ["maya"]
     families = ["colorbleed.renderlayer"]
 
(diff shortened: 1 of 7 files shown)


colorbleed/plugins/maya/publish/increment_current_file_deadline.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def process(self, context):
+
+        from maya import cmds
+
+        from colorbleed.action import get_errored_plugins_from_data
+
+        plugins = get_errored_plugins_from_data(context)
+
+        if any(plugin.__name__ == "MindbenderSubmitDeadline"
```
BigRoy: Did you test whether this works? Maybe once force the `submit_to_deadline` plug-in to raise an error and see whether this is picked up here.

colorbleed/plugins/maya/publish/increment_current_file_deadline.py
```diff
@@ -0,0 +1,75 @@
...
+                for plugin in plugins):
+            raise RuntimeError("Skipping incrementing current file because"
+                               "submission to deadline failed.")
+
+        new_filename = self.version_up(context.data["currentFile"])
+
+        cmds.file(rename=new_filename)
+        cmds.file(save=True, force=True, type="mayaAscii")
+
+    def version_up(self, filepath):
```
BigRoy: Note for selves: I'm expecting this kind of versioning logic to be available through the library or api so we can use it consistently in multiple locations. :)

PR #22, opened 2017-10-24
Implement new-style loaders and fix PLN-26

BigRoy:
Woohoo. :)

- All loaders are now new-style with their own `load`, `update` and `remove` method.
- Fixed PLN-26

This is working with the **latest!** Avalon core, so we'd need to update along with this one. Let's put this live?

### DIFF
--- a/colorbleed/maya/plugin.py
+++ b/colorbleed/maya/plugin.py
@@ -1,0 +1,124 @@
+from avalon import api
+
+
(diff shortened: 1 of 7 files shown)


aardschok:
Fixed a conflict ;) most of all ... let's go!

BigRoy:
Exciting. That messed up some things. :')

PR #24, opened 2017-10-24
Implement new-style loaders and fix PLN-26

BigRoy:
Woohoo. :)

- All loaders are now new-style with their own load, update and remove method.
- Fixed PLN-26
- Fixed mergey :')

This is working with the latest! Avalon core, so we'd need to update along with this one. Let's put this live?


BigRoy:
Also fixed REN-8 just now and an issue when V-Ray image extension was never changed from `.png` and Deadline didn't get the extension right.

PR #26, opened 2017-10-26
Ren 0011

aardschok:
* Created publish imagesequence module to contain the logic
* Added Python Script Job for publishing post render
    * Dependent on render job


aardschok:
Resolving conflict might damage master, closing

PR #34, opened 2017-10-30
Use .py instead of .pyc

aardschok:
Fix REN-18
* Check extension of imported module
* Swap extension to ".py"


BigRoy:
Looking good. I like the second more - good second commit! It's even more explicit if you only ever swap it from the end of the file instead of doing a full replace method.

```python
module_path = publish_imagesequence.__file__[:-len(".pyc")] + ".py"
```

That way it could _never_ happen that it replaces a potential .py that is somewhere else in the filepath.

PR #37, opened 2017-10-31
Add node id relationship comparison

aardschok:
* Solves ANM 07
* Compare deformed shapes with original from history chain
* Added repair action for check
* Added `collect_animation` module to unify collection of hierarchy for both families


colorbleed/plugins/maya/publish/collect_animation.py
```diff
@@ -0,0 +1,63 @@
+import pyblish.api
+
+import maya.cmds as cmds
+
+
+class CollectLook(pyblish.api.InstancePlugin):
```
BigRoy: Voila, this should be `CollectAnimation`, or `CollectionAnimationData`, or whatever. :)

Also update the docstring. :)

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -18,72 +17,116 @@ class ValidateAnimationOutSetRelatedNodeIds(pyblish.api.InstancePlugin):
...
-
-    @classmethod
     def get_invalid(cls, instance):
-        invalid_items = []
+        """Get all nodes which do not match the criteria"""
+        invalid = []
 
         # get asset id
-        nodes = cls.get_pointcache_nodes(instance)
+        nodes = instance.data["pointcache_data"]
```
BigRoy:
```python
instance.data.get("out_hierarchy", instance[:])
```

If we do it like this we don't need to do the collect animation for the _pointcache_ family, only for those that override it. We can even make this more explicit, like `pointcache_override` or `out_data`.

colorbleed/plugins/maya/publish/validate_single_assembly.py
```diff
@@ -3,10 +3,19 @@
 
 
 class ValidateSingleAssembly(pyblish.api.InstancePlugin):
-    """Ensure all nodes are in a single assembly
+    """Ensure the instance is present in the assemblies
```
BigRoy: The new docstring here makes _less sense_ to me.

> Ensure the instance is present in the assemblies

This isn't true. The node in the instance are *not required* to be an assembly, they don't need to be a top node. They just need to be _all together_ in a single contained hierarchy.

Could you simplify this docstring in a way that it's simpler?

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -18,72 +17,116 @@ class ValidateAnimationOutSetRelatedNodeIds(pyblish.api.InstancePlugin):
...
 
     @staticmethod
-    def to_item(id):
+    def to_item(_id):
         """Split the item id part from a node id"""
-        return id.rsplit(":", 1)[0]
+        return _id.split(":", 1)[0]
+
+    @staticmethod
+    def get_parent(node):
```
BigRoy: Let's not make this a `staticmethod` but just make it a separate function at the top to simplify things.

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -18,72 +17,116 @@ class ValidateAnimationOutSetRelatedNodeIds(pyblish.api.InstancePlugin):
...
+        # Check all of the remaining similar nodes and take the first one
+        # with an id and assume it's the original.
+        for similar_node in similar_nodes:
+
+            _id = lib.get_id(similar_node)
+            if not _id:
+                continue
+
+            asset_id = cls.to_item(_id)
+            break
```
BigRoy: Why break here and not just return? :)

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -18,72 +17,116 @@ class ValidateAnimationOutSetRelatedNodeIds(pyblish.api.InstancePlugin):
...
+    @classmethod
+    def repair(cls, instance):
+
+        for node in cls.get_invalid(instance):
+            # Get node ID and the asset ID part
+            node_id = lib.get_id(node)
+            asset_id = cls.to_item(node_id)
+
+            # Get root asset ID
+            root_id = cls.get_history_root_id(node=node)
```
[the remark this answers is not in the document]
aardschok: Because we want the node to have an unique ID which starts with the asset ID of the related asset.
If the Colorbleed ID is exactly the same we will probably run in to trouble with IDs Not Unique.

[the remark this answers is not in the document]
aardschok:
```python
class ValidateNodeIDs(pyblish.api.InstancePlugin):
    """Validate nodes have a Colorbleed Id.
    
    When IDs are missing from nodes *save your scene* and they should be 
    automatically generated because IDs are created on non-referenced nodes 
    in Maya upon scene save.

    """

    order = colorbleed.api.ValidatePipelineOrder
    label = 'Instance Nodes Have ID'
    hosts = ['maya']
    families = ["colorbleed.model",
                "colorbleed.pointcache",
                "colorbleed.look",
                "colorbleed.rig"]
```

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+        str, full path if parent node
+    """
+    return cmds.listRelatives(node, parent=True, fullPath=True)
 
-    An ID is 'related' if its built in the current Item.
 
-    Note that this doesn't ensure it's from the current Task. An ID created
-    from `lookdev` has the same relation to the Item as one coming from others,
-    like `rigging` or `modeling`.
+class ValidateAnimationOutSetRelatedNodeIds(pyblish.api.InstancePlugin):
```
BigRoy: We can shorten this class name. `ValidateOutRelatedNodeIds` or something like that should suffice.

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+            node_type = cmds.nodeType(node)
             node_id = lib.get_id(node)
+
+            if node_type == "mesh" and not node_id:
+                invalid.append(node)
+                continue
+
             if not node_id:
-                invalid_items.append(node)
+                continue
```
BigRoy: I was wondering - what happens if a node does *not* have an id? Does it currently pass all validations? Or which one is still holding it back?

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+        invalid = []
 
         # get asset id
-        nodes = cls.get_pointcache_nodes(instance)
+        nodes = instance.data.get("out_hierarchy", instance[:])
         for node in nodes:
+            node_type = cmds.nodeType(node)
             node_id = lib.get_id(node)
+
+            if node_type == "mesh" and not node_id:
```
BigRoy: This is explicit, but I'm not sure it's safe enough. Aren't there more types than just meshes that require ids? Or is this full validator only *ever* checking meshes? - If it's meshes only should the class and docstring maybe match that.

Looking at the class and docstring I believe it's a general safety net nodes having their id?
Looking at the code it seems to be more a safety net for "deformed meshes" missing their id.

I'm not sure what the intentions are anymore. :)

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+
+        Returns:
+            str: the asset ID as found in the database
+
+        """
+
+        node = cmds.ls(node, long=True)[0]
+
+        # We only check when the node is *not* referenced
+        if cmds.referenceQuery(node, isNodeReferenced=True):
```
BigRoy: Shouldn't this logic of skipping to check this specific node be in `get_invalid` just like you're skipping others there too?

colorbleed/plugins/maya/publish/validate_single_assembly.py
```diff
@@ -3,10 +3,16 @@
 
 
 class ValidateSingleAssembly(pyblish.api.InstancePlugin):
-    """Ensure all nodes are in a single assembly
-
-    Published assets must be contained within a single transform
-    at the root of your outliner.
+    """Ensure the content of the instance is grouped in a single hierarchy
+
+    The instance must have a node which resides within the a group which
```
BigRoy: > the _a_ group

Minor typo?

colorbleed/plugins/maya/publish/collect_animation.py
```diff
@@ -0,0 +1,53 @@
+import pyblish.api
+
+import maya.cmds as cmds
+
+
+class CollectAnimationOutputGeometry(pyblish.api.InstancePlugin):
+    """Collect out hierarchy data for instance.
```
BigRoy: Love this! 👍

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+                invalid.append(node)
+                continue
+
             if not node_id:
-                invalid_items.append(node)
+                continue
+
+            root_id = cls.get_history_root_id(node=node)
+            if root_id is not None:
+                asset_id = cls.to_item(node_id)
```
BigRoy: This line should be removed. Otherwise the following line will always be `True` because it's checking against a non-splitted id. The `to_item` method could then also be removed completely.

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
aardschok: So move this check to `get_invalid` instead of `get_history_root_id`?

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
aardschok: We have the plugin `ValidateNodeIDs` for that, if it needs an ID by that logic it will be caught anyway.

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
aardschok: I have made it less explicit but still pretty explicit with `obj_type = cmds.objectType(node, isAType="shape")` we want to check the shapes because the shapes get the shader ;)

BigRoy: > I have made it less explicit but still pretty explicit with obj_type = cmds.objectType(node, isAType="shape") we want to check the shapes because the shapes get the shader ;)

Sure, I'm fine with explicit. The problem is more that the class/docstring wasn't described as one that is *so explicit*. With this type check the validation is *not* generic at all, it only considers shapes. If we go down that route I'd say it even needs to be "deformable shapes" and the validation is intended to only check for "lost ids" because of deformers.

BigRoy: Also note that a parent constraint is also a "shape", I think deformers are to. Maybe this type check now is to broad for Maya to be accurate?

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
BigRoy: To me that makes more sense, yes. And the `get_history_root_id` could even be a generic function, just like `get_parent`. It doesn't require any information of the class or the instance - in a way it's simple and elegant.

colorbleed/maya/lib.py
```diff
@@ -625,7 +625,7 @@ def _node_type_exists(node_type):
     if _node_type_exists("ilrBakeLayer"):
         ignore |= set(cmds.ls(type="ilrBakeLayer", long=True))
 
-    # establish set of nodes to ignore
+    # Establish set of nodes to include
```
BigRoy: node types? :) but this is fine already.

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+    return cmds.listRelatives(node, parent=True, fullPath=True)
+
+
+class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin):
+    """Validate if deformed shapes have related IDs to the original shapes
 
-    Note that this doesn't ensure it's from the current Task. An ID created
-    from `lookdev` has the same relation to the Item as one coming from others,
-    like `rigging` or `modeling`.
+    Any intermediate shapes which are created when creating deformers on
```
BigRoy:
```
When a deformer is applied in the scene on a referenced mesh that already had
deformers then Maya will create a new shape node for the mesh that does not
have the original id. This validator checks whether the ids are valid on all the shape
nodes in the instance.
```

Does that make more sense as docstring?

aardschok: It explains it a bit more, I will but it as docstrings

colorbleed/plugins/maya/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -3,87 +3,127 @@
...
+            # Check if node is a shape as deformers only work on shapes
+            obj_type = cmds.objectType(node, isAType="shape")
+            if not obj_type:
+                continue
+
+            # Get the current id of the node
             node_id = lib.get_id(node)
-            if not node_id:
-                invalid_items.append(node)
+            if node_id:
```
BigRoy: I don't believe this is what you want @aardschok . Should this be ìf not node_id``?

BigRoy:
Refactored some extra things. :) Maybe run this once in a test environment and see if you have any other issues. But this is looking good.

Note that this *also!* refactors the generate ids and set id logic - maybe a big change for a PR like this. But if we can merge soon I believe this is good, but definitely check the code and see if you have any feedback or preferred changes.

PR #39, opened 2017-11-03
Implement ANM-9

BigRoy:
This code implements ANM-9 by customizing the Maya ui button callbacks (after launch!) to custom callbacks in which we apply special CTRL+click behavior.

It's a bit hacky, but the other alternative is to edit maya's interface `.mel` files which have to be updated and managed for each Maya version. This customize code was written in a way where it's compatible with Maya across multiple versions, hopefully also those to come. Tested this with Maya 2016 and 2018.

@aardschok what you think?

### DIFF
--- a/colorbleed/maya/customize.py
+++ b/colorbleed/maya/customize.py
@@ -16,20 +16,21 @@ def override_component_mask_commands():
 
     This implements special behavior for Maya's component
     mask menu items where a ctrl+click will instantly make
-    it a isolated behavior disabling all others.
+    it an isolated behavior disabling all others.
     
-    Tested in Maya 2016 and 2018.1
+    Tested in Maya 2016 and 2018
 
     """
     log.info("Installing override_component_mask_commands..")
 
-    BUTTONS = mc.formLayout("objectMaskIcons",
+    # Get all object mask buttons
+    buttons = mc.formLayout("objectMaskIcons",
                             query=True,
                             childArray=True)
     # Skip the triangle list item
-    BUTTONS = [btn for btn in BUTTONS if btn != "objPickMenuLayout"]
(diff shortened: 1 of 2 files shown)


colorbleed/maya/customize.py
```diff
@@ -0,0 +1,66 @@
...
+    log.info("Installing override_component_mask_commands..")
+
+    # Get all object mask buttons
+    buttons = mc.formLayout("objectMaskIcons",
+                            query=True,
+                            childArray=True)
+    # Skip the triangle list item
+    buttons = [btn for btn in buttons if btn != "objPickMenuLayout"]
+
+    def _on_changed_callback(raw_command, state):
```
aardschok: This is a private function AND a protected function. Use either a private function or protected function.
I would stay away from nesting functions in functions because, IMO, it makes the code less clear.

BigRoy: It's nested because it's local callback. I don't want it globally. It should stay there.

BigRoy: We could remove the underscore though if that makes it more readable. :)

aardschok: I think your argument for nesting it could have been better. After a minor google search I found this:
> A closure simply causes the inner function to remember the state of its environment when called. Beginners often think that a closure is the inner function, when it’s really caused by the inner function. The closure “closes” the local variable on the stack and this stays around after the the stack creation has finished executing.

If this effect is why you use it I am fully behind using it as a *nested function* but if not I would recommend moving it and set it private.

BigRoy: Yup, it also uses the parent function's `buttons` list, which obviously is another reason. Good point. I'll remove the underscore and we can merge!

BigRoy:
Should be good now. Feel free to merge.

PR #40, opened 2017-11-06
Fix PLN-40

BigRoy:


colorbleed/maya/plugin.py
```diff
@@ -71,6 +71,13 @@ def update(self, container, representation):
         assert os.path.exists(path), "%s does not exist." % path
         cmds.file(path, loadReference=reference_node, type=file_type)
 
+        # Fix PLN-40 for older containers created with Avalon that had the
+        # `.verticesOnlySet` set to True.
+        if cmds.getAttr(container['objectName'] + ".verticesOnlySet"):
```
aardschok: I have a lot of issues with this approach of constantly creating the attribute string.
I suggest you create the string ONCE in a variable and use that one. It will reduce the code to 1 look up in the container dict and the string will only have to be made one instead of twice.

aardschok: https://github.com/BigRoy/colorbleed-config/blob/0999b4b8fa870378d6e7642770d6d01a8f5b2e9f/colorbleed/maya/plugin.py#L55

BigRoy: Sounds good. Didn't see the `node` was already declared. I'll replace them next.

PR #41, opened 2017-11-06
Implement draft version for PLN-10: notify outdated containers w/ pop-up

BigRoy:
This is a working but draft implementation for PLN-10. 

I'm not to happy by the amount of required code for this but the usability is alright - also note the popup widget that I added for this. I'm expecting this pop-up or a similar method of highlighting that something is outdated will become used across multiple hosts, e.g. Fusion. That's why I chose *not* to show it somewhere *in* Maya's interface.

This will show a pop-up parented to the Maya host in the bottom right corner after scene open, like this:
popup

- Pressing the "show" button will open the scene inventory and thus highlight what's outdated. 
- Pressing the "X" or escape will close the popup.

Let's discuss.

---

This will require: https://github.com/getavalon/core/pull/296

### DIFF
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -1,0 +1,47 @@
+import avalon.io as io
+import avalon.api
+
+
+def is_latest(representation):
+    """Return whether the representation is from latest version
+
+    Args:
+        representation (str or io.ObjectId): The representation id.
+
+    Returns:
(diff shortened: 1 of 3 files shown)


BigRoy:
Let's release a prototype of this tomorrow @aardschok. If you have feedback let's see if we can implement it and get this ball rolling. It's getting more and more important to have this in production.

PR #42, opened 2017-11-07
Implement improved logging for `lib.apply_shaders` and cosmetics (PEP08)

BigRoy:
Improve logging for `lib.apply_shaders`. It will now more clearly display why it's not assigning correctly (no shader found or more than one shader found). Also when this happens the errors are logged and the assignment will continue for the rest of the shaders. This way at least all valid shaders are actually assigned instead of failing on the first invalid one.

Also there's some cosmetics cleanup for PEP08. :)


BigRoy:
With https://github.com/Colorbleed/colorbleed-config/pull/42/commits/107c5d0129151ad52725a0691eeb42f14fcb1f43 this also fixes the intermediate objects being checked in looks and models, etc.

aardschok:
Looks good, lets merge it

PR #43, opened 2017-11-07
Improve Validate Unique ID logic

aardschok:
Fixes issue LKD-0008

* Updated Regenerat IDs action for plugins
* Lookdev instance will contain shadingEngines
    * This will enable us to keep using the same approach for all families
* Removed redundant methods
* Simplified code for several validators

### DIFF
--- a/colorbleed/action.py
+++ b/colorbleed/action.py
@@ -1,8 +1,6 @@
 # absolute_import is needed to counter the `module has no cmds error` in Maya
 from __future__ import absolute_import
 
-import uuid
-
 from maya import cmds
 import pyblish.api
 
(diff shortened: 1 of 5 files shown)


colorbleed/action.py
```diff
@@ -197,6 +195,5 @@ def _update_id_attribute(self, instance, nodes):
         asset = instance.data['asset']
         asset_id = io.find_one({"name": asset, "type": "asset"},
                                projection={"_id": True})['_id']
-        for node in nodes:
-            lib.set_id(node, asset_id, overwrite=True)
-
+        for node, _id in lib.generate_ids(nodes, asset_id=asset_id):
```
BigRoy: Good catch! 👍

BigRoy:
Great work! This looks clean. Let's merge it.

PR #44, opened 2017-11-08
Implementation RenderGlobals

aardschok:
Use renderGlobals to:
* Trigger collection of render layers
* Suspend depending publishing job
* Include masterLayer (defaultRenderLayer)
* Override frame range
* Create machine limit (including list type; blacklist / whitelist)


colorbleed/plugins/maya/create/colorbleed_renderglobals.py
```diff
@@ -0,0 +1,49 @@
...
+    def process(self):
+
+        exists = cmds.ls(self.name)
+        assert len(exists) <= 1, (
+            "More than one renderglobal exists, this is a bug")
+
+        if exists:
+            return cmds.warning("%s already exists." % exists[0])
+
+        super(CreateRenderGlobals, self).process()
```
BigRoy: Maybe after this lock the machinelist attribute, just like the tool does?

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -62,43 +80,50 @@ def process(self, context):
...
-                _globals = maya.read(avalon_globals)
-                data["renderGlobals"] = self.get_global_overrides(_globals)
+            # Get global overrides and translate to Deadline values
+            overrides = self.translate_overrides(render_globals)
+            data.update(**overrides)
 
             instance = context.create_instance(layername)
             instance.data.update(data)
 
+            pprint.pprint(instance.data)
```
BigRoy: Is this debugging?

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -62,43 +78,44 @@ def process(self, context):
...
             instance = context.create_instance(layername)
             instance.data.update(data)
 
     def get_render_attribute(self, attr):
         return cmds.getAttr("defaultRenderGlobals.{}".format(attr))
 
-    def get_global_overrides(self, globals):
-        """
-        Get all overrides with a value, skip those without
+    def translate_overrides(self, render_globals):
```
BigRoy: Maybe this reads better when called `parse_options`? or `parse_overrides`?

BigRoy: Here's another draft that might inspire to reformat some code:
```python
    def parse_options(self, render_globals):
        """Get all overrides with a value, skip those without

        Here's the kicker. These globals override defaults in the submission
        integrator, but an empty value means no overriding is made.
        Otherwise, Frames would override the default frames set under globals.

        Args:
            render_globals (str): collection of render globals

        Returns:
            dict: only overrides with values
        """

        attributes = maya.read(render_globals)

        options = {"renderGlobals": {}}
        
        options['renderGlobals']['Priority'] = attributes['priority']
        
        # Machine list
        machine_list = attributes["machineList"]
        if machine_list:
            key = "Whitelist" if  attributes["whitelist"] else "Blacklist"
            options['renderGlobals'][key] = machine_list
        
        # Suspend publish job
        state = "Suspended" if attributes["suspendPublishJob"] else "Active" 
        options["suspendPublishJob"] = state 

        return options
```

colorbleed/plugins/maya/publish/submit_deadline.py
```diff
@@ -194,7 +195,10 @@ def process(self, instance):
         })
 
         # Include optional render globals
-        payload["JobInfo"].update(instance.data.get("renderGlobals", {}))
+        render_globals = instance.data.get("renderGlobals", {})
+        payload["JobInfo"].update(render_globals)
+
+        pprint.pprint(payload["JobInfo"])
```
BigRoy: Debugging?

PR #51, opened 2017-11-22
Update `context.empty_sets`

aardschok:
This pull requests needs to be done after the `cb` has been updated with the latest PR
* Updated `context.empty_sets` call, add `force` flag
* Reduced dictionary calls for `look data relationships`


BigRoy:
Looking good - let's merge!

> Reduced dictionary calls for look data relationships

Note that this is purely an optimization that would only "feel" faster if this was called millions of times - readability should remain top priority here. No need to go down those lengths if at this point they somewhat reduce readability. :) The previous felt as if it was directly transferring the lookdata, the new might feel like it's a relationships data that has been altered or created locally. ;)

PR #54, opened 2017-11-28
Fix REN-23

aardschok:
* Ensure referenced render layers are collected


BigRoy:
> Ensure referenced render layers are collected

Actually this does the **exact oppossite**! 

This disregards referenced renderlayers because they are incompatible with the `maya.cmds.editRenderLayerGlobals` command that switches towards the layer - which cannot be done for referenced layers (one can't switch to them). 

To fix the collector it now ignores any referenced renderlayers.

PR #55, opened 2017-11-29
Introduce Yeti cache & rig 

aardschok:
Introduces Yeti Rig and Yeti Cache publishing and loading:
* New family type : yetiRig and yeticache
* New collectors and extractors

Future:
* Loader for Yeti Rig will be supported by a tool to manage complex connections

### DIFF
--- a/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
+++ b/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
@@ -1,0 +1,15 @@
+import avalon.maya
+
+
+class CreateYetiRig(avalon.maya.Creator):
+    """Output for procedural plugin nodes ( Yeti / XGen / etc)"""
+
+    name = "yetiDefault"
+    label = "Yeti Cache"
+    family = "colorbleed.yeticache"
+    icon = "pagelines"
+
+    def __init__(self, *args, **kwargs):
+        super(CreateYetiRig, self).__init__(*args, **kwargs)
+
+        self.data["peroll"] = 0
--- a/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
+++ b/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
@@ -1,4 +1,7 @@
+from collections import OrderedDict
+
 import avalon.maya
+from colorbleed.maya import lib
 
 
 class CreateYetiRig(avalon.maya.Creator):
@@ -12,4 +15,11 @@ class CreateYetiRig(avalon.maya.Creator):
     def __init__(self, *args, **kwargs):
         super(CreateYetiRig, self).__init__(*args, **kwargs)
 
-        self.data["peroll"] = 0
+        data = OrderedDict(self.data)
+        data["peroll"] = 0
+
+        anim_data = lib.collect_animation_data()
+        data.update({"startFrame": anim_data["startFrame"],
+                     "endFrame": anim_data["endFrame"]})
+
+        self.data = data
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -1,0 +1,17 @@
+from avalon import api
+
+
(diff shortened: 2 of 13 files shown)


colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
```diff
@@ -0,0 +1,25 @@
...
+    icon = "pagelines"
+
+    def __init__(self, *args, **kwargs):
+        super(CreateYetiRig, self).__init__(*args, **kwargs)
+
+        data = OrderedDict(self.data)
+        data["peroll"] = 0
+
+        anim_data = lib.collect_animation_data()
+        data.update({"startFrame": anim_data["startFrame"],
```
BigRoy: Should "handles" be important too? (So a shot going from 1 to 70 with handles of 1 also has motion blur on frame 70 for example.)

colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -0,0 +1,166 @@
...
+            # Create new pgYetiMaya node
+            node_name = "{}:{}".format(namespace, node)
+            yeti_node = cmds.createNode("pgYetiMaya", name=node_name)
+            cmds.connectAttr("time1.outTime", "%s.currentTime" % yeti_node)
+
+            # Apply colorbleed ID to node
+            lib.set_id(node=yeti_node,
+                       unique_id=settings["cbId"],
+                       overwrite=True)
+            settings.pop("cbId", None)
```
BigRoy: The fact that `cbId` uses such custom logic compared with the other settings I'm thinking should this even be just in the settings? Should this be more explicitly referred to as the node's id?

For example:

```python
{"id": "xx/yy",
 "attribute": {"renderWidth": 1.0, 
                   "renderDensity": 2.0"}
}
```

Or does that make less sense?

colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -0,0 +1,166 @@
...
+
+        try:
+            cmds.delete(nodes)
+        except ValueError:
+            # Already implicitly deleted by Maya upon removing reference
+            pass
+
+        cmds.namespace(removeNamespace=namespace, deleteNamespaceContent=True)
+
+    def update(self, container, representation):
```
BigRoy: Note that updating logic eventually will become more complex. For example in these scenarios:

- What if a pgYetiMaya node in the source file was removed or added?
- What if the render density or render width in the source file was changed? Should this override it again? What if the local scene had "overrides"?

---

And a little note to self, the `.fur` representation is a folder, so this `api.get_representation_path(representation)` will return a folder, not a file - which is fine.

colorbleed/plugins/publish/integrate.py
```diff
@@ -162,18 +168,25 @@ def register(self, instance):
...
                 )
 
+                assert not any(os.path.isabs(name) for name in collection)
+
                 template_data["representation"] = ext[1:]
 
                 for fname in collection:
-                    src = os.path.join(stagingdir, fname)
+
+                    if os.path.isabs(fname):
```
BigRoy: This is redundant with the assertion above

colorbleed/plugins/maya/publish/validate_yeti_rig_inputs.py
```diff
@@ -0,0 +1,40 @@
+from maya import cmds
+
+import pyblish.api
+import colorbleed.api
+
+
+class ValidateRigInputMeshes(pyblish.api.Validator):
```
BigRoy: Should this be `ValidateYetiRigInputMeshes`?

BigRoy:
@aardschok You can consider some of these comments (or maybe all) as **TODO** - we're currently prototyping the Yeti things in production and I'd prefer to just iterate a lot on this. Get a draft out into the hands of the artists ; they basically need it *now!*.

So let's get this live if it's functional and we can always force them to "republish" if that's required. It'll likely break here and there - but I think we just want to start having more people testing this.

PR #57, opened 2017-11-30
Added attribute change uuid check

aardschok:
Ensure to check if the node of a custom attribute / changed attribute has a UUID


BigRoy:
Looking good. Note that this might end up being *too* restrictive - but let's use it in production and see what happens. :)

PR #58, opened 2017-11-30
Updated Yeti Rig / Cache Loader 

aardschok:
* Resolved issue with Manager showing correct version of updated Yeti Cach
* Removed `post_process` logic from Yeti Rig Loader (to be replaced by tool)

### DIFF
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -124,13 +124,18 @@ class YetiCacheLoader(api.Loader):
 
         path = api.get_representation_path(representation)
         members = cmds.sets(container['objectName'], query=True)
-        yeti_node = cmds.ls(members, type="pgYetiMaya", long=True)
+        all_members = cmds.listRelatives(members, ad=True)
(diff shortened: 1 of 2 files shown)


colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -124,14 +124,19 @@ def update(self, container, representation):
 
         path = api.get_representation_path(representation)
         members = cmds.sets(container['objectName'], query=True)
-        yeti_node = cmds.ls(members, type="pgYetiMaya", long=True)
+        all_members = cmds.listRelatives(members, ad=True)
```
BigRoy: Actually, wait.. no! This means the problem *must!* be elsewhere - the container should always contain all content!

aardschok: > Actually, wait.. no! This means the problem must! be elsewhere - the container should always contain all content!

When a Yeti node''s transform node, similar to a mesh's transfrom node, is added we don't get the `pgYetiMaya` or `pgYetiGroom` nodes. In hindsight you are correct, this will mean that the loading of representation into the scene is incorrect. I will improve this asap!

PR #59, opened 2017-11-30
Verbosity update

aardschok:
Ensure verbosity is at all for debug render message


colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -83,6 +83,8 @@ def load(self, context, name=None, namespace=None, data=None):
                          image_search_path,
                          type="string")
 
+            cmds.setAttr("%s.verbosity" % yeti_node, 2)
```
BigRoy: Could you elaborate on why we need this?

PR #62, opened 2017-12-01
Updated collect data & related logic

aardschok:
Yeti:
* Added transform information (node ID and name)
* Attributes of shape node are stored in their own key,  `"attrs": {attr: value, attr: value}`
* Updated loading logic to match new data structure

Look:
* Collect all user attributes of nodes with ID which are not referenced within an instance

This will allow the current approach of loading Yeti caches in lookdev to have objectIDs for compositing / rendering.

### DIFF
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -47,66 +47,12 @@ class YetiCacheLoader(api.Loader):
 
         # Get node name from JSON
         nodes = []
-        node_data = fursettings["nodes"]
-        for node_settings in node_data:
-
-            # Create transform node
-            transform = node_settings["transform"]
-            transform_node = cmds.createNode("transform",
-                                             name=transform["name"])
-
-            lib.set_id(transform_node, transform["cbId"])
-
-            # Create pgYetiMaya node
-            original_node = node_settings["name"]
-            node_name = "{}:{}".format(namespace, original_node)
-            yeti_node = cmds.createNode("pgYetiMaya",
-                                        name=node_name,
-                                        parent=transform_node)
-
-            lib.set_id(yeti_node, node_settings["cbId"])
-
-            nodes.append(transform_node)
-            nodes.append(yeti_node)
-
-            # Apply attributes to pgYetiMaya node
-            kwargs = {}
-            for attr, value in node_settings["attrs"].items():
-                attribute = "%s.%s" % (yeti_node, attr)
-                if isinstance(value, (str, unicode)):
-                    cmds.setAttr(attribute, value, type="string")
-                    continue
-                cmds.setAttr(attribute, value, **kwargs)
-
-            # Ensure the node has no namespace identifiers
-            node_name = original_node.replace(":", "_")
-
-            # Create full cache path
-            cache = os.path.join(self.fname, "{}.%04d.fur".format(node_name))
-            cache = os.path.normpath(cache)
-            cache_fname = self.validate_cache(cache)
-            cache_path = os.path.join(self.fname, cache_fname)
-
-            # Preset the viewport density
-            cmds.setAttr("%s.viewportDensity" % yeti_node, 0.1)
-
-            # Add filename to `cacheFileName` attribute
-            cmds.setAttr("%s.cacheFileName" % yeti_node,
-                         cache_path,
-                         type="string")
-
-            cmds.setAttr("%s.imageSearchPath" % yeti_node,
-                         image_search_path,
-                         type="string")
-
-            # Set verbosity for debug purposes
-            cmds.setAttr("%s.verbosity" % yeti_node, 2)
-
-            # Enable the cache by setting the file mode
-            cmds.setAttr("%s.fileMode" % yeti_node, 1)
-
-            nodes.append(yeti_node)
-            nodes.append(transform_node)
+        if "nodes" in fursettings:
+            node_data = fursettings["nodes"]
+            self.create_nodes(namespace, image_search_path, node_data)
+        else:
+            # Backwards compatibilty
+            self.create_nodes_old(namespace, image_search_path, fursettings)
 
         group_name = "{}:{}".format(namespace, asset["name"])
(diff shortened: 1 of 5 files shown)


BigRoy:
Not sure if this code was already there, but I don't feel so comfortable with the `socketId` terminology here: https://github.com/aardschok/colorbleed-config/blob/40b68f9839c570581f916e69200dd25cea6b3d91/colorbleed/plugins/maya/publish/extract_yeti_rig.py#L22
The same I'd say for the `plugId`. What's a socket? Where is that terminology coming from? And what's a plug?

At first I thought a "socket" would be the one taking in something else, but it's the source output in this code. The same confusion happens with the `plug` where the plug is the one "you're sticking in" but in the code it's other way around - it's the one taking the connection into it. Alongside that I was assuming that with _plug_ you were referring to a maya plug (like a maya node attribute) but you're using that for the "plug-socket" terminology.

Would this make more sense when described as "sourceId" and "destinationId"? That way it's also in line with `connectAttr` and also the `listConnections` method?

colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -46,70 +46,16 @@ def load(self, context, name=None, namespace=None, data=None):
...
-            nodes.append(yeti_node)
-            nodes.append(transform_node)
+        if "nodes" in fursettings:
+            node_data = fursettings["nodes"]
+            nodes = self.create_nodes(namespace, image_search_path, node_data)
+        else:
+            # Backwards compatibilty
+            self.log.info("Encountered old data, "
+                          "using backwards compatibility")
+            nodes = self.create_nodes_old(namespace,
```
BigRoy: Blegh, I so hate this. backwards compatibility; blegh!

Anyhow, would this make more sense to "update" the older data to the new one so we could pass it along?

Say you'd find the old data you'd refactor it to match to the new data and apply that instead? That way you wouldn't need the duplicated "old/new" code.

aardschok: Currently the cleaning of the published files is possible and possibly the road of least resistance.
I will refactor the code AFTER I know for sure the updating is correct.

colorbleed/plugins/maya/publish/collect_yeti_rig.py
```diff
@@ -37,28 +37,30 @@ def process(self, instance):
                                          fullPath=True) or input_content
 
         # Get all the shapes
-        input_shapes = cmds.ls(input_nodes, long=True)
+        input_shapes = cmds.ls(input_nodes, long=True, noIntermediate=True)
 
         # Store all connections
+        print "input shape:", input_shapes
```
BigRoy: Debug print statement is still in - this should be removed.

colorbleed/plugins/maya/publish/extract_yeti_rig.py
```diff
@@ -17,36 +17,40 @@ def disconnected_attributes(settings, members):
...
         for input in settings["inputs"]:
 
-            # get source
-            socket_id = input["socketID"]
-            sources = lib.lsattr("cbId", socket_id)
-            sources = [i for i in sources if
+            # Get source shapes
+            source_nodes = lib.lsattr("cbId", input["sourceID"])
+            sources = [i for i in source_nodes if
                        not cmds.referenceQuery(i, isNodeReferenced=True)
```
BigRoy: This seems like odd logic for this context manager - why does it require the input to be a referenced node? If that's required than collection must already have required it and it must've been validated if it breaks otherwise.

Furthermore, I'm expecting to always find at most a single match - if there are more; what do we do? Is that an ambiguous error? Do we allow breaking both of them? Or do we want to assert there's always a single match so it's explicit about what connection is broken?

colorbleed/plugins/maya/publish/extract_yeti_rig.py
```diff
@@ -17,36 +17,40 @@ def disconnected_attributes(settings, members):
...
+
+            cmds.disconnectAttr(src_attribute, dst_attribute)
             original_connection.append([src_attribute, dst_attribute])
-            cmds.disconnectAttr(dst_attribute, src_attribute)
         yield
-
     finally:
         # restore connections
         for connection in original_connection:
             src, dest = connection
```
BigRoy: Note that you don't need to do this on an extra line, you can also do:
```python
for src, dest in original_connection:
    pass
```

colorbleed/plugins/maya/publish/extract_yeti_rig.py
```diff
@@ -17,36 +17,40 @@ def disconnected_attributes(settings, members):
...
-            dst_attribute = "%s.%s" % (dst, connections[1])
+            src_attribute = "%s.%s" % (source, connections[0])
+            dst_attribute = "%s.%s" % (destination, connections[1])
 
             # store connection pair
+            if not cmds.isConnected(src_attribute, dst_attribute):
+                continue
+
+            cmds.disconnectAttr(src_attribute, dst_attribute)
             original_connection.append([src_attribute, dst_attribute])
```
BigRoy: Just noticed now that the name `original_connection` for the variable sounds off, it's a list of things. Should that be `original_connections`?

colorbleed/plugins/maya/publish/extract_yeti_rig.py
```diff
@@ -96,9 +100,11 @@ def process(self, instance):
         attr_value = {"%s.imageSearchPath" % n: image_search_path for
                       n in yeti_nodes}
 
-        # get input_SET members
+        # Get input_SET members
         input_set = [i for i in instance if i == "input_SET"]
         members = cmds.sets(input_set[0], query=True)
+        # Get all items
+        members = cmds.listRelatives(members, ad=True, fullPath=True)
```
BigRoy: Actually, are you sure this does exactly what you want? :) Because the contents of the set itself are now skipped, only children are remaining in the data.

If you want to have the contents of the set *and* the children you'd need to somehow merge it with whatever  you passed to `cmds.listRelatives` - preferably in a way that would ensure uniqueness; note that `cmds.sets(query=True)` doesn't enforce long paths.

PR #63, opened 2017-12-06
Implement legacy renderlayer support

aardschok:
* Toggle renderlayer system in renderglobals node for Deadline
* Pass value to renderlayer data

### DIFF
--- a/colorbleed/plugins/maya/publish/submit_deadline.py
+++ b/colorbleed/plugins/maya/publish/submit_deadline.py
@@ -160,6 +160,9 @@ class MindbenderSubmitDeadline(pyblish.api.InstancePlugin):
                 # Only render layers are considered renderable in this pipeline
                 "UsingRenderLayers": True,
 
+                # Use legacy Render Layer system
+                "UseLegacyRenderLayers": True,
+
                 # Render only this layer
                 "RenderLayer": renderlayer,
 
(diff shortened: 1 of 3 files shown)


colorbleed/plugins/maya/publish/submit_deadline.py
```diff
@@ -160,6 +160,9 @@ def process(self, instance):
                 # Only render layers are considered renderable in this pipeline
                 "UsingRenderLayers": True,
 
+                # Use legacy Render Layer system
+                "UseLegacyRenderLayers": True,
```
BigRoy: I think you're missing the usage of the actual instance's setting.

colorbleed/plugins/maya/publish/submit_deadline.py
```diff
@@ -160,6 +160,9 @@ def process(self, instance):
                 # Only render layers are considered renderable in this pipeline
                 "UsingRenderLayers": True,
```
BigRoy: > UsingRenderLayers

I didn't see this one in the plug-in submission data of Deadline itself - what is this supposed to do?

BigRoy: Had a quick look, it is being used. In our case it isn't actually doing anything because it's always used like this:

```
if self.Renderer == "file" or ( self.UsingRenderLayers and len( self.RenderLayer ) == 0 ):
```

And our renderer is always set to to `"file"` in our submissions - but for now just let's keep the code in. :)

PR #66, opened 2017-12-07
Update Yeti pipeline

aardschok:
* Force connect Yeti shape node to time node
* Fixed updating version
* Collect all set members (set members and their relatives)
* Cosmetics

### DIFF
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -195,7 +195,7 @@ class YetiCacheLoader(api.Loader):
             cmds.setAttr("%s.fileMode" % yeti_node, 1)
 
             # Connect to the time node
-            cmds.connectAttr("time1.timeOut", "%s.currentTime" % yeti_node)
+            cmds.connectAttr("time1.outTime", "%s.currentTime" % yeti_node)
 
             nodes.append(yeti_node)
             nodes.append(transform_node)
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -46,7 +46,7 @@ class YetiCacheLoader(api.Loader):
         node_data = fursettings["nodes"]
         nodes = self.create_nodes(namespace, node_data)
 
-        group_name = "{}:{}".format(namespace, asset["name"])
+        group_name = "{}:{}".format(namespace, name)
         group_node = cmds.group(nodes, name=group_name)
 
         nodes.append(group_node)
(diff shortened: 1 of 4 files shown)


colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -84,15 +84,26 @@ def remove(self, container):
...
+        yeti_node = cmds.ls(members, type="pgYetiMaya")
+
+        # TODO: Count the amount of nodes cached
+        # To ensure new nodes get created or old nodes get destroyed
 
         for node in yeti_node:
-            node_name = node.split(":")[-1]
+            # Remove local given namespace
+            node_name = node.split(namespace, 1)[-1]
+            node_name = node_name.replace(":", "_")
```
BigRoy: What `:` will remain present after the namespace was removed? What is this replacing?

Also, what's the expected idea if magically the namespace does not match? Do we need to assert? (just wondering).

colorbleed/plugins/maya/publish/extract_yeti_rig.py
```diff
@@ -102,9 +101,10 @@ def process(self, instance):
 
         # Get input_SET members
         input_set = [i for i in instance if i == "input_SET"]
-        members = cmds.sets(input_set[0], query=True)
         # Get all items
-        members = cmds.listRelatives(members, ad=True, fullPath=True)
+        set_members = cmds.sets(input_set[0], query=True)
+        members = cmds.listRelatives(set_members, ad=True, fullPath=True)
```
BigRoy: Should we for now also implement the `or []` fallback for list relatives; otherwise this would error if `input_SET` has no members at all.

BigRoy:
Going into a great direction. Looking good. I think some things still don't feel to secure - like the updating, but for current production usage it'll be fine. We'll assume for now a rig will never increase/decrease/rename any of the yeti nodes. :)

aardschok:
> We'll assume for now a rig will never increase/decrease/rename any of the yeti nodes. :)

Completely true, I am thinking to use as source of the amount of nodes needed. We can validate the amount of nodes in container X are enough or some need to be created or deleted. A bit saver approach i.m.o., what is your take on it?

colorbleed/plugins/maya/load/load_yeti_cache.py
aardschok: When a cache is generated in a shot the Yeti node's full name is `character_x_01_:body_hairShape`  for example. When using the <NAME> token will take the namespace and node name as the full node name. When loading those caches in to the shot the name will be `shot0010:character_x_01_:body_hairShape`. Hence the split and replace, it is a quirk I will need to figure out.

The namespace is based on the namespace from the container. An assertion is possible but I do not see any danger in not asserting at the moment.

PR #67, opened 2017-12-08
Fix updating of referenced content when .placeHolderList attributes are in container

BigRoy:
- Fix updating of referenced content when .placeHolderList attributes are present in the container.
- This update will also remove any .placeHolderList members when updating since they are redundant data for the container.


BigRoy:
Merging this - let me know if you've any questions or comments on implementation and we can patch that up later.

PR #81, opened 2018-01-08
Solves issue REN-29

aardschok:
* Group instances based on family not name
* Sort render layers based on the `displayOrder` attribute ( display order as seen in Render Layer Manager )
* Added forgotten select action to `ValidateSingleShader`

### DIFF
--- a/colorbleed/plugins/maya/publish/collect_instances.py
+++ b/colorbleed/plugins/maya/publish/collect_instances.py
@@ -113,11 +113,12 @@ class CollectInstances(pyblish.api.ContextPlugin):
             # user interface interested in visualising it.
             self.log.info("Found: \"%s\" " % instance.data["name"])
 
+        def sort_by_family(instance):
+            """Sort by family"""
+            return instance.data.get("families", instance.data.get("family"))
+
         # Sort/grouped by family (preserving local index)
-        grouped = sorted(enumerate(context), key=self.sorter)
-        context[:] = [x[1] for x in grouped]
-        # context[:] = sorted(context, key= lambda x: (x.data['families'],
-        #                                              x.data['name']))
+        context[:] = sorted(context, key=sort_by_family)
 
         return context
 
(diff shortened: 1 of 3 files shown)


colorbleed/plugins/maya/publish/collect_instances.py
```diff
@@ -113,7 +113,11 @@ def process(self, context):
             # user interface interested in visualising it.
             self.log.info("Found: \"%s\" " % instance.data["name"])
 
-        context[:] = sorted(context)
+        # Sort/grouped by family (preserving local index)
+        grouped = sorted(enumerate(context), key=self.sorter)
+        context[:] = [x[1] for x in grouped]
+        # context[:] = sorted(context, key= lambda x: (x.data['families'],
```
BigRoy: Remove the commented lines :)

Also, I personally wouldn't push the `sorted` method to the class because it's so specific to these specific lines of code. Which is related to just replacing a lambda in PEP08, see:

- https://www.python.org/dev/peps/pep-0008/#programming-recommendations
- https://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

These can and are recommended even to be inline if they are just that, like a lambda.

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -39,6 +38,11 @@ def process(self, context):
                         cmds.getAttr("{}.renderable".format(i)) and not
                         cmds.referenceQuery(i, isNodeReferenced=True)]
 
+        # Sort by displayOrder
+        renderlayers = sorted(renderlayers,
+                              key=lambda x: -cmds.getAttr("%s.displayOrder" % x))
+
+        renderlayers = reversed(renderlayers)
```
BigRoy: Actually this is double? Did you add this reversed? Because the `negative` sign in front of displayOrder should already reverse it. :)

PR #84, opened 2018-01-25
Fixed PLN-68

aardschok:
Fixes issue with unable to containerize loaded camera because the `cmds.camera` function does not support `lockTransform` flag. This flag has been introduced in Maya 2017

### DIFF
--- a/colorbleed/plugins/maya/load/load_camera.py
+++ b/colorbleed/plugins/maya/load/load_camera.py
@@ -27,8 +27,15 @@ class CameraLoader(colorbleed.maya.plugin.ReferenceLoader):
                           returnNewNodes=True)
 
         cameras = cmds.ls(nodes, type="camera")
-        for camera in cameras:
-            cmds.camera(camera, edit=True, lockTransform=True)
+
+        # Check the Maya version, lockTransform has been introduced since
+        # Maya 2017
+        if cmds.about(version=True) >= "2018":
+            for camera in cameras:
+                cmds.camera(camera, edit=True, lockTransform=True)
+        else:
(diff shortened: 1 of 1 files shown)


colorbleed/plugins/maya/load/load_camera.py
```diff
@@ -27,8 +27,15 @@ def process_reference(self, context, name, namespace, data):
                           returnNewNodes=True)
 
         cameras = cmds.ls(nodes, type="camera")
-        for camera in cameras:
-            cmds.camera(camera, edit=True, lockTransform=True)
+
+        # Check the Maya version, lockTransform has been introduced since
+        # Maya 2017
```
BigRoy: It's funny how you say it's introduced in Maya 2017 but then force the check to only allow Maya 2018 and up. haha. :)

colorbleed/plugins/maya/load/load_camera.py
```diff
@@ -27,8 +27,15 @@ def process_reference(self, context, name, namespace, data):
...
+        if cmds.about(version=True) >= "2018":
```
BigRoy: Oof, the comparison on strings. Will we rely on this? I believe maya 2016 Extension 2 actually return "2016.5" which will be (in a string comparison) larger than "2018".

Should we force it to an integer?
```
    maya_version = int(cmds.about(version=True))
    if maya_version >= 2018:
        # do whateverrrrr
        pass
```

aardschok:
Fixes can be found here:
`https://github.com/Colorbleed/colorbleed-config/pull/85`

PR #87, opened 2018-02-12
PLN 0053

aardschok:
Solved issue where the input meshes were being validated but checked the wrong direction, checked source instead of destination.
Refactored `glob` out of the collector


colorbleed/plugins/maya/publish/validate_yeti_rig_inputs.py
```diff
@@ -24,15 +28,26 @@ def get_invalid(cls, instance):
...
 
-        input_set = [i for i in instance if i == "input_SET"][0]
-        input_nodes = cmds.sets(input_set, query=True)
-        for node in input_nodes:
-            incoming = cmds.listConnections(node,
-                                            source=True,
-                                            destination=False,
-                                            connections=True,
-                                            plugs=True)
+        input_set = [i for i in instance if i == "input_SET"]
```
BigRoy: Here `input_set` ends up being a list. Maybe this makes more sense as:
`input_set = next(i for i in instance if i == "input_SET")`

If you want to additionally assert instead of error.

```python
input_set = next(i for i in instance if i == "input_SET", None)
assert input_set, "Derpy"
```

aardschok:
Changed the validator to a more logic one as the input mesh can have a lot of connections. If we want to check those I will need to rethink the strategy all together. I think you even previously said it would not be needed to check or too complex to check, something along those lines.

PR #90, opened 2018-02-22
Fusion integration

BigRoy:
Implement Fusion functionality.

A first working configuration for Fusion! :+1: 

This supports loading and managing of image sequences, and also publishing savers to both Deadline and Local renders. 

#### Some Maya changes
It's important to note that this also includes some cleanup, cosmetics (among others [1](https://github.com/Colorbleed/colorbleed-config/commit/ea43cdd868643a26c4e00677d3b67e8f8659b775) and [2](https://github.com/Colorbleed/colorbleed-config/commit/c143f164d0ad06fd7b115a3e8cc381208d5b0795)) and code changes to Maya functionality, like supporting [multiple UV sets](https://github.com/Colorbleed/colorbleed-config/commit/e97fa88a121ccc13d1dfe091e9bdd5b6048af903), [support of locking of camera in maya 2017](https://github.com/Colorbleed/colorbleed-config/commit/4ab6b530e366fdb9d9399be53a84fae19e842b67), a large simplification of the publish imagesequence script and interface.


colorbleed/plugins/fusion/create/create_saver.py
```diff
@@ -0,0 +1,46 @@
+import os
```
aardschok: Note for future selfs, change the module name: too generic

colorbleed/plugins/fusion/load/actions.py
```diff
@@ -0,0 +1,103 @@
...
+        comp.SetAttrs(attrs)
+
+
+class FusionSetFrameRangeLoader(api.Loader):
+    """Specific loader of Alembic for the avalon.animation family"""
+
+    families = ["colorbleed.animation",
+                "colorbleed.camera",
+                "colorbleed.imagesequence",
+                "colorbleed.yeticache",
```
aardschok: Yeti Cache has no place in Fusion, point cache is for future implementation if it ever is available in the API to load.

BigRoy: Ah, no - this isn't actually loading the cache. It solely allows you to "take" the frame range from the yeti cache and apply it to the fusion scene. It's a very generic action, it should support anything that has a `startFrame` and `endFrame` data. So this is actually correct.

colorbleed/plugins/fusion/create/create_saver.py
BigRoy: Thanks - fixed.

colorbleed/plugins/fusion/load/actions.py
BigRoy: Also note, pointcache is actually used in production. :+1:

aardschok: Yes it is used in production but I am assuming you do not need to line up your comp length based on a pointcache which is rendered to an image sequence which is loaded in your comp? :thinking:

PR #91, opened 2018-03-09
Introduce switching in config

aardschok:
* Added simple switch UI for in Fusion
* Added fusion.lib module
* Added switch_item to colorbleed.lib
* Updated Fusion actions accordingly
* Added switch methods to all loaders
* Added fusion_switch_shot module (backbone of logic for UI)


colorbleed/scripts/fusion_switch_shot.py
```diff
@@ -0,0 +1,213 @@
...
+    Args:
+        comp (object): current comp instance
+        asset (dict): asset document of the asset to update TO
+
+    Returns:
+         None
+    """
+
+    new_work = _get_work_folder(session)
+    # TODO
```
BigRoy: This doesn't seem finished. :) (also see the line below it with the question mark, hehe)

BigRoy:
Other than above comment it looks good. :muscle:

colorbleed/scripts/fusion_switch_shot.py
```diff
@@ -33,7 +33,7 @@ def _format_version_folder(folder):
     if not os.path.isdir(folder):
         return "v001"
 
-    re_version = re.compile("v\d+")
+    re_version = re.compile("v\d+$")
 
     content = sorted(os.listdir(folder))
```
BigRoy: Not to be a bitch, note that - in this specific scenario - there's no reason to actually sort the output of `os.listdir`. Not really an issue though. :smile:

aardschok: Oh darn that's true :) probably as a result of assuming there would be only v### folder but I changed my mind about that.

BigRoy:
If you've tested this and it works - let's merge! :)

aardschok:
Found an issue with checking for the version folder after the merge. 

> What if there are no version folders in the "renders" folder?

I will setup a `FUS-29b` branch to ensure this is fixed

PR #95, opened 2018-03-22
Implement slapcomp workflow

aardschok:
Main points
* Adds __fusion/scripts/deadline_submit_and_switch.py__
* Deadline support through BlackmagicFusion module with use of Fusion Console Node
* Plugin for creation of switch job 

-Note: part of issue FUS-32-


aardschok:
Saw an issues with retrieving the `listedslaves`, will fix this before merging

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -135,4 +138,10 @@ def parse_options(self, render_globals):
         state = "Suspended" if attributes["suspendPublishJob"] else "Active"
         options["suspendPublishJob"] = state
 
+        # Check if the run slap comp
+        if attributes["runSlapComp"]:
```
BigRoy: This will break backwards compatibility for render global nodes that don't have the slap comp attribute yet, right?

colorbleed/plugins/maya/publish/submit_switch_job.py
```diff
@@ -0,0 +1,108 @@
...
+
+        filepath = instance.data("flowFile", "")
+        if not filepath:
+            raise RuntimeError("No flow file (comp) chosen")
+
+        shot = api.Session["AVALON_ASSET"]
+        comment = instance.context.data["comment"]
+
+        scriptdir = _get_script_dir()
+        scriptfile = os.path.join(scriptdir, "deadline_swith_and_submit.py")
```
BigRoy: Did you test this? :) Probably won't work because of a typo. The filename is: `deadline_switch_and_submit.py`

aardschok: > Saw an issues with retrieving the listedslaves, will fix this before merging

I saw this as well when attempting to fix the above mentioned issue

colorbleed/plugins/maya/publish/submit_switch_job.py
```diff
@@ -0,0 +1,108 @@
...
+            raise Exception(response.text)
+
+        # Temporary key name, deadlineSubmissionJob was already taken
+        if instance.data("runSlapComp", False):
+            instance.data["deadlineDependJob"] = response.json()
+
+        self.log.info("Slap comp arguments: %s" % args)
+
+    def _get_list_type(self, job):
+        return "Whitelist" if job["Props"]["White"]is True else "Blacklist"
```
BigRoy: Missing space in `job["Props"]["White"]is` - it's interesting that this is actually valid python code.

colorbleed/plugins/publish/submit_publish_job.py
```diff
@@ -159,3 +159,7 @@ def process(self, instance):
         response = requests.post(url, json=payload)
         if not response.ok:
             raise Exception(response.text)
+
+        # Temporary key name, deadlineSubmissionJob was already taken
+        if instance.data("runSlapComp", False):
```
BigRoy: I don't think this is doing what you intended. Probably meant: `instance.data.get("runSlapComp", False)`?

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,202 @@
...
+import sys
+import os
+
+# This script only works with Python 2.7 and 3.6
+version = "{0}{1}".format(*sys.version_info)  # {major}{minor}
+assert version in ["27", "36"], "Script only works in Python 2.7 or 3.6"
+key = "FUSION_PYTHON{0}_HOME".format(version)
+
+# Set Python 3.6 home for fusion, debug
+print("Settings FUSION_PYTHON36_HOME ..")
```
BigRoy: In Python 2.7 this won't be setting `FUSION_PYTHON36_HOME` but `FUSION_PYTHON27_HOME`

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,202 @@
...
+import time
+import sys
+import os
+
+# This script only works with Python 2.7 and 3.6
+version = "{0}{1}".format(*sys.version_info)  # {major}{minor}
+assert version in ["27", "36"], "Script only works in Python 2.7 or 3.6"
+key = "FUSION_PYTHON{0}_HOME".format(version)
+
+# Set Python 3.6 home for fusion, debug
```
BigRoy: This isn't for debugging purposes but is a explicit and direct fix for a crash on `BlackmagicFusion` library crashing.

Let's state that in the comment so the intent is clear.

This could just read:
```python
# Importing BlackmagicFusion package in standalone Python interpreter
# crashes when not installed on default location but runs from, e.g. a
# network share. Forcing Fusion's Python home magically fixes it.
print("Workaround BlackmagicFusion import crash by setting %s" % key)
os.environ[key] = os.path.dirname(sys.executable)
```

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,202 @@
...
+print("Settings FUSION_PYTHON36_HOME ..")
+os.environ[key] = os.path.dirname(sys.executable)
+
+FUSCRIPT_EXE = r"C:/Program Files/Blackmagic Design/Fusion9/FuScript.exe"
```
BigRoy: For me this is fine for now, but should be "defined" elsewhere eventually. Maybe add a TODO line for now?

setup/fusion/scripts/Comp/colorbleed/switch_ui.py
```diff
@@ -154,8 +171,12 @@ def _on_switch(self):
 
         asset = self._assets.currentText()
 
-        import colorbleed.scripts.fusion_switch_shot as switch_shot
-        switch_shot.switch(asset_name=asset, filepath=file_name, new=True)
+        # Get  switch and submit module
+        script_dir = _get_script_dir()
+        site.addsitedir(script_dir)
+
+        import switch_and_submit as switch_shot
```
BigRoy: Importing a script like this seems hacky, kind of feels like the `switch_shot` method is almost a library function of sorts or should just be embedded here in this file. Not sure if there's any use case outside of this file?

It would be much easier if we could just make this file easier accessible then go through all this length to find it and import it. Can't we simplify this?

colorbleed/scripts/fusion/switch_and_submit.py
```diff
@@ -144,22 +146,31 @@ def update_frame_range(comp, representations):
...
 
+    assert asset_name, "Function requires at least an asset name"
+
     # Ensure filename is absolute
-    if not os.path.abspath(filepath):
-        filepath = os.path.abspath(filepath)
+    if file_path:
+        file_path = os.path.abspath(file_path)
+
+    api.install(avalon.fusion)
```
BigRoy: It feels odd to me that this function is installing the host. Shouldn't that be done outside of it - where we actually launch the host?

Hmm... that's just a thought. Seems to me that this function is just doing too much because I was looking at it just as a function one might call to switch whatever is in the current scene, but it's doing much more. It's opening comps, publishing jobs, etc. It's probably a remainder of it originally being the script itself.

Currently it's more a `load_comp_switch_asset_and_submit` kind of method which I'm not "expecting" from the function name and docstring.

setup/fusion/scripts/Comp/colorbleed/switch_ui.py
```diff
@@ -146,16 +148,20 @@ def _refresh(self):
...
     def _on_switch(self):
 
+        asset = self._assets.currentText()
+        _comp = avalon.fusion.get_current_comp()
         if not self._use_current.isChecked():
             file_name = self._comps.itemData(self._comps.currentIndex())
+            # Get the current comp's App (FusionUI)
+            _fusion = _comp.GetApp()
+            # Open the selected comp
+            _fusion.LoadComp(file_name)
```
BigRoy: This might fail (e.g. file is invalid), ensure it returns the newly loaded comp.

setup/fusion/scripts/Comp/colorbleed/switch_ui.py
```diff
@@ -146,16 +148,20 @@ def _refresh(self):
...
         else:
-            comp = avalon.fusion.get_current_comp()
-            file_name = comp.GetAttrs("COMPS_FileName")
-
-        asset = self._assets.currentText()
+            file_name = _comp.GetAttrs("COMPS_FileName")
 
-        import colorbleed.scripts.fusion_switch_shot as switch_shot
-        switch_shot.switch(asset_name=asset, filepath=file_name, new=True)
+        switched = flib.switch(asset)
```
BigRoy: Hmm, this wasn't directly clear to me as to what it returns. Is `switched_comp` more readable?

setup/fusion/scripts/Comp/colorbleed/switch_ui.py
aardschok: This won't be the case as the filename is derived from browsing to the file, see `_on_open_from_dir`

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,221 @@
...
+    # Set comp render mode to deadline
+    current_comp.SetData("colorbleed.rendermode", "deadline")
+
+    error_format = "Failed {plugin.__name__}: {error} -- {error.traceback}"
+
+    # Publish
+    context = api.publish()
+    if not context:
+        log.warning("Nothing collected.")
+        sys.exit(1)
```
BigRoy: Shouldn't this also raise an error instead of hard exiting? Maybe the logged message could just be the raised message.

```python
raise RuntimeError("Nothing collected for publish")
```

colorbleed/fusion/lib.py
```diff
@@ -38,3 +45,154 @@ def update_frame_range(start, end, comp=None, set_render_range=True):
...
+
+    # Build the session to switch to
+    switch_to_session = api.Session.copy()
+    switch_to_session["AVALON_ASSET"] = asset['name']
+
+    # Update session and environment
+    api.Session.update(switch_to_session)
+    os.environ.update(switch_to_session)
+
+    self._project = None
```
BigRoy: Ok, so the `self._project` is only for an optimization local to this function. How about allowing to pass the `project` as argument to the methods that require them? That way it's much more explicit and also avoids the redundant complexity of storing it globally. What do you think?

aardschok: I am all up for it, no objections.

setup/fusion/scripts/Comp/colorbleed/switch_ui.py
BigRoy: > This won't be the case as the filename is derived from browsing to the file, see _on_open_from_dir

The file could change after browsing, plus the file itself could be in invalid working comp. The result will be that the switch command will be switching the wrong comp around - which is dangerous. Right?

aardschok: > plus the file itself could be in invalid working comp

Define an invalid working comp?

I can add `os.path.isfile` to ensure the file is not deleted in the meantime. 

>Furthermore, if the file is invalid it will show a popup message too! ;) Which you might want to suppress?

I would like Fusion to give me a pop up message if the comp file is invalid, I can add a filter to find only `.comp` files,

BigRoy: > I would like Fusion to give me a pop up message if the comp file is invalid, I can add a filter to find only .comp files,

Save an empty file name `test.comp`, try to open it.

```python
fusion.LoadComp("C:/test.comp")
print("A")
```

The "A" will be printed after you've clicked the pop-up. It will continue onwards in your script running on the wrong composition.

I'm testing with an empty file, but it could just as well be any comp file that didn't save correctly.

colorbleed/scripts/GlobalJobPreLoad.py
```diff
@@ -0,0 +1,42 @@
+import os
+import site
+
+# todo: refactore out the need of site
```
BigRoy: This needs to be solved prior to merge too.

colorbleed/scripts/launch.py
```diff
@@ -0,0 +1,57 @@
...
+    tools_env = api.get_tools(tools.split(";"))
+    env = api.compute(tools_env)
+
+    env = api.merge(env, current_env=dict(os.environ))
+    print("Environment:\n%s" % pprint.pformat(env, indent=4))
+
+    # Search for the executable within the tool's environment
+    # by temporarily taking on its `PATH` settings
+    original = os.environ["PATH"]
+    os.environ["PATH"] = env.get("PATH", os.environ.get("PATH", ""))
```
BigRoy: Let's also get this workaround out, just add in the extra argument for lib.which :)

colorbleed/fusion/lib.py
```diff
@@ -38,3 +42,118 @@ def update_frame_range(start, end, comp=None, set_render_range=True):
...
+                                                    asset_name=asset_name)
+            representations.append(representation)
+            log.debug(str(representation["_id"]) + "\n")
+        except Exception as e:
+            log.debug("Error in switching! %s\n" % e.message)
+
+    log.info("Switched %i Loaders of the %i\n" % (len(representations),
+                                                  len(containers)))
+    # Updating frame range
+    log.debug("\nUpdating frame range ..")
```
BigRoy: If Fusion requires the `\n` then this one is ordered differently then the other logs above ;) The new line should then be at the end?

Maybe we can find a way to fix it in the Fusion logger, so it doesn't need the new line explicitly here. (Magic!)

colorbleed/launcher_actions.py
```diff
@@ -0,0 +1,44 @@
...
+        Returns:
+            Popen instance of newly spawned process
+
+        """
+
+        # Update environment with session
+        env = os.environ.copy()
+        env.update(session)
+
+        # Get executable by na.e
```
BigRoy: Typo in comment, `name`

colorbleed/launcher_actions.py
```diff
@@ -0,0 +1,44 @@
...
+        app = lib.get_application(self.name)
+        executable = lib.which(app["executable"])
+
+        return lib.launch(executable=executable, args=[], environment=env)
```
BigRoy: The `env` does not include anything of an environment from the `app` - so this doesn't actually include the app's environment? That seems wrong.

colorbleed/scripts/launch.py
aardschok: Can we replace this module with the one from our `env_prototype` ( we need to pick that awesome name ;) )

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,310 @@
...
+
+import traceback
+import logging
+import time
+import site
+import sys
+import os
+
+
+site.addsitedir(r"P:\pipeline\dev\git\env_prototype")
```
BigRoy: Remove :)

colorbleed/scripts/fusion/deadline_switch_and_submit.py
```diff
@@ -0,0 +1,310 @@
...
+    # Start a fusion console node in "listen" mode
+    tools_env = acre.get_tools(["global", "fusionnode9"])
+    env = acre.compute(tools_env)
+    env = acre.merge(env, current_env=dict(os.environ))
+
+    # Search for the executable within the tool's environment
+    # by temporarily taking on its `PATH` settings
+    original = os.environ["PATH"]
+    os.environ["PATH"] = env.get("PATH", os.environ.get("PATH", ""))
+    exe = lib.which("fusionconsolenode")
```
BigRoy: This could use the `acre.which`? So you don't need to use the hack?

colorbleed/scripts/launch.py
```diff
@@ -0,0 +1,45 @@
+import os
+import pprint
+
+import acre
+
+
+config_root = os.path.dirname(os.path.dirname(__file__))
+os.environ["TOOL_ENV"] = os.path.join(config_root, "environments")
```
BigRoy: This is not correct? Since we have a standalone repo for the environments.

colorbleed/scripts/launch.py
```diff
@@ -0,0 +1,45 @@
...
+    print("Environment:\n%s" % pprint.pformat(env, indent=4))
+
+    # Search for the executable within the tool's environment
+    # by temporarily taking on its `PATH` settings
+    exe = acre.which(executable, env)
+    if not exe:
+        raise ValueError("Unable to find executable: %s" % executable)
+
+    print("Launching: %s" % exe)
+    acre.launch(exe, environment=env, args=args, cwd=env.get("AVALON_WORKDIR"))
```
BigRoy: Wasn't this full file's functionality replaced by acre's launching?

PR #96, opened 2018-03-23
Introduce ValidateNodeIdsInDatabase

aardschok:
Solves PLN-88

* Check all id required nodes of an instance against the database to ensure the asset id in the `cbId` attribute is an existsing asset in the current project.


aardschok:
Seems something went wrong, branch should only have one file added. Closing this now

PR #97, opened 2018-03-23
Introduce ValidateNodeIdsInDatabase

aardschok:
Solved PLN-88

Retrieve all id required nodes' their asset id and compare them with the registered asset ids in the database

### DIFF
--- a/colorbleed/plugins/maya/publish/validate_node_ids_in_database.py
+++ b/colorbleed/plugins/maya/publish/validate_node_ids_in_database.py
@@ -49,7 +49,7 @@ class ValidateNodeIdsInDatabase(pyblish.api.InstancePlugin):
             cb_id = lib.get_id(node)
             asset_id = cb_id.split(":", 1)[0]
             if asset_id not in db_asset_ids:
-                cls.log.error("`%s` has unassociated assetfd ID" % node)
+                cls.log.error("`%s` has unassociated asset ID" % node)
                 invalid.append(node)
 
         return invalid
--- a/colorbleed/plugins/maya/publish/validate_node_ids_in_database.py
+++ b/colorbleed/plugins/maya/publish/validate_node_ids_in_database.py
@@ -1,0 +1,55 @@
+import pyblish.api
+
+import avalon.io as io
+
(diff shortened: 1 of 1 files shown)


colorbleed/plugins/maya/publish/validate_node_ids_in_database.py
```diff
@@ -0,0 +1,55 @@
...
+        # check ids against database ids
+        db_asset_ids = io.find({"type": "asset"}).distinct("_id")
+        db_asset_ids = set(str(i) for i in db_asset_ids)
+
+        # Get all asset IDs
+        for node in id_required_nodes:
+            cb_id = lib.get_id(node)
+            asset_id = cb_id.split(":", 1)[0]
+            if asset_id not in db_asset_ids:
+                cls.log.error("`%s` has unassociated assetfd ID" % node)
```
BigRoy: Little typo in `assetfd`

BigRoy:
Nice work. Looks good other than the little typo.

BigRoy:
Question: shouldn't it just check all found node ids, instead of only the required node ids? That way _any ids found_ won't be wrong either?

aardschok:
It's a reasonable question but with the current pipeline we do not care about every node, hence the `id_required_nodes` function ;) So nodes with an 'cbId' attribute which are not in the id required nodes can safely be ignored and should not impact the pipeline. 

If you want to check all nodes per instance I would not name the plugin `ValidateNodeIdsInDatabase ` but a more general name as `ValidateNodesPipelineCompatible` or create that plugin alongside `ValidateNodeIdsInDatabase`. 

In `ValidateNodesPipelineCompatible` we check all nodes against the `id_required_nodes` and if there are node which have a `cbId` attribute but are not part of the required nodes we mark them as invalid. This way we ensure they don't have extra `cbId` attribute on nodes which do not matter for the current pipeline. On the other hand ... what about other custom attributes. Generally we do not care about them.

BigRoy:
Sure, just wanted to have it noted. With this nodes may still have ids that are not coming from the current project - which might end up being confusing.

Looks good!

PR #100, opened 2018-03-28
Add FusionRenderNode action

aardschok:
* Added `register_launcher_actions`
* Added FusionRenderNode class


BigRoy:
Awesome, this is a perfect start!

1. Move the actions somewhere more globally, since they are registered globally?

Hmm, it's a bit odd to import the `colorbleed.fusion` in the `__init__` to register the actions. Should we maybe just have an `actions.py` that contains all actions. Especially because the Action doesn't actually require Fusion logic - it just happens to start a Fusion application. So should we move this?

I noticed there's already an `action.py` containing different Actions (those for pyblish) so we need to somehow ensure there's little confusion about what is what.

2. The Fusion render node is unfinished.

The Fusion render node looks like good code, it's just that we'll still need to configure it so it has the environment similar to the "Fusion" application as if it were run within the pipeline, like the OFX plug-ins from the server etc. For that to be operational we'll need to ensure it mimics that environment correctly. Currently it's this one: `"P:\pipeline\2.1_dev\git\avalon-setup\bin\fusion9.toml"` that is used to run Fusion with the correct environment.

aardschok:
Both points have been addressed.

1. has been moved to `config_actions`, I do not 100% stand behind the naming so if can provided a better name I am all ear.

2. I have implemented an approach similar to other application actions to launch this in the correct environment with the correct plugins etc.

```python
        # Update environment with session
        env = os.environ.copy()
        env.update(session)

        # Get executable by na.e
        app = lib.get_application(self.name)
        executable = lib.which(app["executable"])
```

colorbleed/__init__.py
```diff
@@ -1,6 +1,7 @@
 import os
 from pyblish import api as pyblish
-from avalon import api, pipeline
+
+from .config_actions import register_actions
```
BigRoy: If you refactor the `register_actions` to `register_launcher_actions` then you don't need the additional function here at lines 22-24. Then the import is just there to expose it to the launcher. :) Better?

BigRoy:
> has been moved to config_actions, I do not 100% stand behind the naming so if can provided a better name I am all ear

`launcher_actions` maybe more explicit?

colorbleed/__init__.py
BigRoy: Not sure why this shows as outdated as I commented on line 4. :confused:

aardschok: Nice catch! Will make things a bit clearer :)

BigRoy:
AWESOME!

PR #102, opened 2018-04-04
Implement PLN-103: Refactor ImportModelLoader to more generic ImportMayaLoader

BigRoy:
> Note: Previously loaded containers with ImportModelLoader cannot be managed anymore with this PR because of the refactored name of the Loader.


BigRoy:
Merging this, still let me know if you think another way is better @aardschok

PR #107, opened 2018-04-13
Fix PLN-108 & PLN-109

BigRoy:
Fixes read errors on referenced files and also is more explicit about which reference node to use from the container, it now uses the highest level reference node usually the top node, but could - with this code - still work when the container is also inside a reference and thus is not a top reference node.


BigRoy:
Grazi! :smile:

PR #111, opened 2018-04-24
Re-initialize Maya after context changed

aardschok:
* Fixes issue PLN55 ( temporarily till )
* Adds wrapped function of context changed logic and existing on task changed function
* Override taskChanged event callback with wrapped function


BigRoy:
Quirky and funky workaround, but I guess it does it for now. :)

We need to remain aware that any callbacks hooked to `taskChanged` will get removed at that point of time - even other callbacks we might register to it.

If we need this in for production, feel free to merge.

PR #113, opened 2018-04-25
Add custom action for manager

aardschok:
* Solves FUS040 in combination with [PR19](https://github.com/Colorbleed/core/pull/19)
* Select related tool of selected entry
* Specified module for Manager actions

### DIFF
--- a/colorbleed/__init__.py
+++ b/colorbleed/__init__.py
@@ -3,7 +3,6 @@ from pyblish import api as pyblish
 from avalon import api as avalon
 
 from .launcher_actions import register_launcher_actions
-from .manager_actions import register_manager_actions
 
 PACKAGE_DIR = os.path.dirname(__file__)
 PLUGINS_DIR = os.path.join(PACKAGE_DIR, "plugins")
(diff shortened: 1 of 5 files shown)


colorbleed/fusion/__init__.py
```diff
@@ -20,6 +22,8 @@ def install():
 
     pyblish.register_callback("instanceToggled", on_pyblish_instance_toggled)
 
+    colorbleed.register_manager_actions()
```
BigRoy: No need to take this from the colorbleed global space. The fusion actions should just belong to the fusion host - install it like any of the other actions.

Maybe it should even be added as files in:

```plugins/fusion/inventory```

:+1:

File: colorbleed/__init__.py [hunk 2]
```diff
@@ -3,6 +3,7 @@
 from avalon import api as avalon
 
 from .launcher_actions import register_launcher_actions
+from .manager_actions import register_manager_actions
```
BigRoy: Let's see if we can drop this here. :) Since we don't need the global register_manager_actions - see other comment.

colorbleed/fusion/__init__.py
aardschok: Nice! I like that idea :)

BigRoy:
Nice, I think artists would love a similar action for Maya - let's start working on that too.

Other than that I think this looks great!

I love how little is necessary on this end to get this working.

PR #115, opened 2018-05-03
Introduce fetching of additional data

aardschok:
The `collect_container_metadata` functions allows for custom metadata to be collected in each host through the use of the `registered_config` and `registered_host` function of Avalon api.
Each host in the config will have its own `get_additional_data` to ensure it is relevant only for that host.

This merge will introduce the usability of this logic for Fusion and is accompanied with an Inventory plugin.
Please node that this plugin will not work unless [this PR](https://github.com/Colorbleed/core/pull/19) has been resolved together with [this PR](https://github.com/Colorbleed/core/pull/20) for the `SceneInventory`

### DIFF
--- a/colorbleed/__init__.py
+++ b/colorbleed/__init__.py
@@ -1,4 +1,6 @@
 import os
+import importlib
+
 from pyblish import api as pyblish
 from avalon import api as avalon
 
@@ -23,3 +25,36 @@ def uninstall():
     pyblish.deregister_plugin_path(PUBLISH_PATH)
     avalon.deregister_plugin_path(avalon.Loader, LOAD_PATH)
 
+
+def _get_host_name():
+
+    _host = avalon.registered_host()
+    # This covers nested module name like avalon.maya
+    return _host.__name__.rsplit(".", 1)[-1]
+
+
+def collect_container_metadata(container):
+    """Add additional data based on the current host
+
+    If the host application's lib module does not have a function to inject
+    additional data it will return the input container
+
+    Args:
+        container (dict): collection if representation data in host
+
+    Returns:
+        generator
+    """
+
+    host_name = _get_host_name()
+
+    # This will cover nested module names like avalon.maya
+    package_name = "{}.{}.lib".format(__name__, host_name)
+    hostlib = importlib.import_module(package_name)
+
+    if not hasattr(hostlib, "get_additional_data"):
+        print("{} has no function called "
+              "get_additional_data".format(package_name))
+        return container
+
+    return hostlib.get_additional_data(container)
--- a/colorbleed/__init__.py
+++ b/colorbleed/__init__.py
@@ -1,5 +1,4 @@
 import os
-import importlib
 
 from pyblish import api as pyblish
 from avalon import api as avalon
@@ -24,37 +23,3 @@ def uninstall():
     print("Deregistering global plug-ins..")
     pyblish.deregister_plugin_path(PUBLISH_PATH)
     avalon.deregister_plugin_path(avalon.Loader, LOAD_PATH)
-
-
-def _get_host_name():
-
-    _host = avalon.registered_host()
-    # This covers nested module name like avalon.maya
-    return _host.__name__.rsplit(".", 1)[-1]
-
-
-def collect_container_metadata(container):
-    """Add additional data based on the current host
-
-    If the host application's lib module does not have a function to inject
-    additional data it will return the input container
-
-    Args:
-        container (dict): collection if representation data in host
-
-    Returns:
-        generator
-    """
-
-    host_name = _get_host_name()
-
-    # This will cover nested module names like avalon.maya
-    package_name = "{}.{}.lib".format(__name__, host_name)
-    hostlib = importlib.import_module(package_name)
-
-    if not hasattr(hostlib, "get_additional_data"):
-        print("{} has no function called "
-              "get_additional_data".format(package_name))
-        return container
-
-    return hostlib.get_additional_data(container)
--- a/colorbleed/fusion/lib.py
+++ b/colorbleed/fusion/lib.py
@@ -1,5 +1,6 @@
 import sys
 
+from avalon.vendor.Qt import QtGui
 import avalon.fusion
 
 
@@ -58,7 +59,6 @@ def get_additional_data(container):
     if tile_color is None:
         return container
 
-    tile_color.pop("__flags", None)
-    rgb = {key: clamp(value) for key, value in tile_color.items()}
-
-    return {"color": "#{R:02x}{G:02x}{B:02x}".format(**rgb)}
+    return {"color": QtGui.QColor(clamp(tile_color["R"]),
+                                  clamp(tile_color["G"]),
+                                  clamp(tile_color["B"]))}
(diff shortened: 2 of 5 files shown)


colorbleed/__init__.py
```diff
@@ -23,3 +25,36 @@ def uninstall():
...
+    host_name = _get_host_name()
+
+    # This will cover nested module names like avalon.maya
+    package_name = "{}.{}.lib".format(__name__, host_name)
+    hostlib = importlib.import_module(package_name)
+
+    if not hasattr(hostlib, "get_additional_data"):
+        print("{} has no function called "
+              "get_additional_data".format(package_name))
+        return container
```
BigRoy: This shouldn't return the container - correct? It should then return an empty dictionary?

colorbleed/fusion/lib.py
```diff
@@ -38,3 +39,26 @@ def update_frame_range(start, end, comp=None, set_render_range=True):
...
+        dict
+    """
+
+    def clamp(value):
+        return int(value * 255)
+
+    tool = container["_tool"]
+    tile_color = tool.TileColor
+    if tile_color is None:
+        return container
```
BigRoy: This should not return the container in that case, but an empty dictionary.

colorbleed/plugins/fusion/inventory/select_loader_color.py
```diff
@@ -0,0 +1,40 @@
+from avalon import api
+from avalon.vendor.Qt import QtGui, QtWidgets
+
+import avalon.fusion
+
+
+class FusionSelectLoaderColor(api.InventoryAction):
+    """Update the color of the selected tools"""
+
+    label = "Select Loader Color"
```
BigRoy: Let's refactor this to **Set Tool Color** - also use "FusionSetToolColor" for the top action, because if additional containers are to be supported it should end up working for those as well.

colorbleed/plugins/fusion/inventory/select_loader_color.py
```diff
@@ -0,0 +1,40 @@
...
+    icon = "plus"
+    color = "#d8d8d8"
+
+    def process(self, containers):
+        """Color all selected tools the selected colors"""
+
+        comp = avalon.fusion.get_current_comp()
+
+        # Launch pick color
+        color = QtGui.QColor(1.0, 1.0, 1.0)
```
BigRoy: Can we have the "start color" in the UI start with the currently set color, for example from the first container or the last. Just seems odd that it starts with a pure white even if you want to update the color of a single container.

Or what's possible in that regard?

colorbleed/plugins/fusion/inventory/select_loader_color.py
```diff
@@ -0,0 +1,40 @@
...
+                rgb_f = picked_color.getRgbF()
+                rgb_f_table = {"R": rgb_f[0], "G": rgb_f[1], "B": rgb_f[2]}
+
+                # Update tool
+                tool = container["_tool"]
+                tool.TileColor = rgb_f_table
+
+        self.refresh()
+
+    def refresh(self):
```
BigRoy: Ah, interesting! Actually... looking at this. Whenever an action is perform in the Scene Inventory, shouldn't it always update afterwards right in the Scene Inventory? Let's see if we can remove this here - otherwise I'm expecting we'll start having hacks like this in many actions.

colorbleed/__init__.py
```diff
@@ -23,3 +25,36 @@ def uninstall():
     pyblish.deregister_plugin_path(PUBLISH_PATH)
     avalon.deregister_plugin_path(avalon.Loader, LOAD_PATH)
 
+
+def _get_host_name():
```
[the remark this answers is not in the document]
aardschok: We could move it to `colorbleed.lib`, I think that is a decent alternative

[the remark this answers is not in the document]
BigRoy: :+1: let's do that for now.

colorbleed/plugins/fusion/inventory/select_loader_color.py
aardschok: Enabling actions to trigger the host's `refresh`  in a non-hack way would be very nice to have :+1:

colorbleed/plugins/fusion/inventory/select_loader_color.py
aardschok: I did have that before but I thought it was a nice thing to have but not really the main goal of the action. Now that I pass on a `QColor` instance it makes things easier.

colorbleed/__init__.py
aardschok: You are correct! If we return a container we would update the new item with the container. Even though that would not be a problem currently it will reduce computing time ;)

aardschok:
Added requested improvements and some more

colorbleed/plugins/fusion/inventory/set_tool_color.py
```diff
@@ -0,0 +1,59 @@
...
+        Returns:
+            QtGui.QColor
+
+        """
+
+        color_dialog = QtWidgets.QColorDialog(color)
+        color_dialog.setStyleSheet(style.load_stylesheet())
+
+        accepted = color_dialog.exec_()
+        picked_color = color_dialog.selectedColor() if accepted else False
```
BigRoy: Does this look nicer?
 
```python
if not accepted:
    return

return color_dialog.selectedColor()
```

It's just cosmetics, so it's up to you. Behavior will be the same.

colorbleed/plugins/fusion/inventory/set_tool_color.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def process(self, containers):
+        """Color all selected tools the selected colors"""
+
+        result = []
+        comp = avalon.fusion.get_current_comp()
+
+        # Launch pick color
+        first = containers[0]
+        color = QtGui.QColor(first.get("color", self._fallback_color))
```
BigRoy: Get the color from the tool, if None fallback to default color.

colorbleed/fusion/lib.py
```diff
@@ -38,3 +39,26 @@ def update_frame_range(start, end, comp=None, set_render_range=True):
...
+
+    def clamp(value):
+        return int(value * 255)
+
+    tool = container["_tool"]
+    tile_color = tool.TileColor
+    if tile_color is None:
+        return {}
+
+    return {"color": QtGui.QColor(clamp(tile_color["R"]),
```
BigRoy: Can't you avoid the `clamp` function and just use [QColor.fromRGBf](http://doc.qt.io/archives/qt-4.8/qcolor.html#fromRgbF)?

colorbleed/lib.py
```diff
@@ -214,3 +215,37 @@ def switch_item(container,
...
+    If the host application's lib module does not have a function to inject
+    additional data it will return the input container
+
+    Args:
+        container (dict): collection if representation data in host
+
+    Returns:
+        generator
+    """
+
```
BigRoy: I think this is fine for now, yet maybe we should add a `# todo` line that states that we should look for simpler/nicer ways to do this?

BigRoy:
Aside from the two comments above this looks great! :+1: nice work.

PR #116, opened 2018-05-07
Added set_family_state for cbloader

aardschok:
In combination with [this PR](https://github.com/Colorbleed/core/pull/21) adds a family filter based on the config. This is to ensure the right family names are toggled on when the loader is initialized.


BigRoy:
Looking much better already :)

PR #123, opened 2018-06-28
Implement project FPS

aardschok:
To improve the flexibility of the projects we added a `get_project_fps` function to check if the host application has the same fps as the project.

In `colorbleed.lib`:
* Get project FPS

In `colorbleed.maya.__init__` and `colorbleed.maya.lib`
* Setting FPS to match project add launch (not all types are covert yet!)
* Setting FPS to match project when opening a file
* Validate on save if FPS matches project

Pyblish Maya plugins:
* Updated validator for Maya Units to check current FPS against project config

### DIFF
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -247,3 +247,23 @@ def collect_container_metadata(container):
         return {}
 
     return hostlib.get_additional_data(container)
+
+
+def get_project_fps():
+    """Returns project's FPS, if not found will return 25 by default
+
+    Returns:
+        int, float
+    """
+
+    project_name = io.active_project()
+    project = io.find_one({"name": project_name,
+                           "type": "project"},
+                          projection={"config.fps": True})
+
+    config = project.get("config", None)
+    assert config, "This is a bug"
+
+    fps = config.get("fps", 25)
+
+    return fps
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -259,7 +259,7 @@ def get_project_fps():
     project_name = io.active_project()
     project = io.find_one({"name": project_name,
                            "type": "project"},
-                          projection={"config.fps": True})
+                          projection={"config": True})
 
     config = project.get("config", None)
     assert config, "This is a bug"
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -251,7 +251,7 @@ def get_project_fps():
     """Returns project's FPS, if not found will return 25 by default
 
     Returns:
-        float
+        int, float
     """
 
     project_name = io.active_project()
@@ -264,4 +264,4 @@ def get_project_fps():
 
     fps = config.get("fps", 25.0)
 
-    return float(fps)
+    return fps
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -253,7 +253,7 @@ def get_project_fps():
     """Returns project's FPS, if not found will return 25 by default
 
     Returns:
-        int, float
+        float
     """
 
     project_name = io.active_project()
@@ -264,6 +264,6 @@ def get_project_fps():
     config = project.get("config", None)
     assert config, "This is a bug"
 
-    fps = config.get("fps", 25)
+    fps = config.get("fps", 25.0)
 
-    return fps
+    return float(fps)
--- a/colorbleed/maya/__init__.py
+++ b/colorbleed/maya/__init__.py
@@ -122,7 +122,7 @@ def on_open(_):
     from ..widgets import popup
 
     # Ensure scene's FPS is set to project config
-    lib.set_project_fps()
+    lib.validate_fps()
 
     # Update current task for the current scene
     update_task_from_path(cmds.file(query=True, sceneName=True))
@@ -174,5 +174,3 @@ def on_task_changed(*args):
 
     # Run
     maya.pipeline._on_task_changed()
-
-
--- a/colorbleed/maya/__init__.py
+++ b/colorbleed/maya/__init__.py
@@ -2,7 +2,7 @@ import os
 import logging
 import weakref
 
-from maya import utils, cmds, mel
+from maya import utils, cmds, mel, OpenMaya as om
 
 from avalon import api as avalon, pipeline, maya
 from pyblish import api as pyblish
@@ -36,6 +36,8 @@ def install():
     avalon.on("init", on_init)
     avalon.on("save", on_save)
     avalon.on("open", on_open)
+
+    avalon.before("save", before_save_check)
 
     log.info("Overriding existing event 'taskChanged'")
     override_event("taskChanged", on_task_changed)
@@ -95,6 +97,11 @@ def on_init(_):
     safe_deferred(override_component_mask_commands)
 
 
+def before_save_check(return_code, _):
+    """Run validation for scene's FPS prior to saving"""
+    return lib.validate_fps()
+
+
 def on_save(_):
     """Automatically add IDs to new nodes
 
@@ -111,8 +118,6 @@ def on_save(_):
     nodes = lib.get_id_required_nodes(referenced_nodes=False)
     for node, new_id in lib.generate_ids(nodes):
         lib.set_id(node, new_id, overwrite=False)
-
-    lib.validate_fps()
 
 
 def on_open(_):
--- a/colorbleed/maya/__init__.py
+++ b/colorbleed/maya/__init__.py
@@ -126,15 +126,17 @@ def on_save(_):
                              QtWidgets.QApplication.topLevelWidgets()}
 
         parent = top_level_widgets.get("MayaWindow", None)
+        if parent is None:
+            pass
+        else:
+            dialog = popup.Popup(parent=parent)
+            dialog.setWindowTitle("Maya scene not in line with project")
+            dialog.setMessage("The FPS is out of sync, please fix")
+            # Set new text for button (add optional argument for the popup?)
+            dialog.widgets["show"].setText("Fix")
+            dialog.on_show.connect(lib.set_project_fps)
 
-        dialog = popup.Popup(parent=parent)
-        dialog.setWindowTitle("Maya scene not in line with project")
-        dialog.setMessage("The FPS is out of sync, please fix")
-        # Set new text for button (could be an optional argument for the popup)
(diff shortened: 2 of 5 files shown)


colorbleed/maya/__init__.py
```diff
@@ -112,13 +112,18 @@ def on_save(_):
...
+
 
 def on_open(_):
     """On scene open let's assume the containers have changed."""
 
     from avalon.vendor.Qt import QtWidgets
     from ..widgets import popup
 
+    # Ensure scene's FPS is set to project config
+    lib.set_project_fps()
```
BigRoy: Don't set the FPS - it'll be too confusing for the artist that somehow the scene just magically changed. It should always just be a clear warning pop-up that states your currently set FPS and what it should be according to the project.

colorbleed/maya/lib.py
```diff
@@ -1292,3 +1312,68 @@ def _get_parent(node):
...
+    int_fps = [15, 24, 5, 30, 48, 50, 60, 44100, 48000]
+    float_fps = [23.976, 29.97, 29.97, 47.952, 59.94]
+
+    fps = lib.get_project_fps()
+
+    if not isinstance(fps, (int, float)):
+        raise ValueError("Set value for project's FPS is not a number. "
+                         "Only accepts floats and integers")
+
+    if int(fps) in int_fps:
```
BigRoy: Doesn't it make more sense to first check whether the number is a float and whether it's in the float list? (Because that's the most exact entry point) -> If not found, check if the integer version is a good match.

For example:
```python
if isinstance(fps, float) and fps in float_fps:
    unit = {:d}fps".format(int(fps))

elif int(fps) in int_fps:
    unit = "{:f}fps".format(fps)

else:
    raise ("Unsupported FPS value: `%s`" % fps)

log.info("Updating FPS to '{}'".format(unit))
cmds.currentUnit(time=unit)
```

Note that this is not a bug in the code, it should be functional because none of the float numbers when using `int(number)`  coincide with anything in the int list. Otherwise if `23`, `29`, `47` or `59` would be in the float list it would never match any of those in the float list, because it would always match the one from the int list.

colorbleed/maya/lib.py
```diff
@@ -1292,3 +1312,68 @@ def _get_parent(node):
...
+
+def set_project_fps():
+    """Set FPS from project configuration
+
+    Returns:
+        None
+
+    """
+
+    int_fps = [15, 24, 5, 30, 48, 50, 60, 44100, 48000]
```
BigRoy: Can we make these sets instead of list for optimal lookup (faster)

```python
int_fps = {15, 24, 5, 30, 48, 50, 60, 44100, 48000}
float_fps = {23.976, 29.97, 29.97, 47.952, 59.94}
```

colorbleed/maya/lib.py
aardschok: Thank you for the pointer

BigRoy: Note that my code block isn't entirely correct. The `unit = xx` lines are swapped by accident. :)

colorbleed/maya/lib.py
```diff
@@ -1322,27 +1322,22 @@ def set_project_fps():
...
+    int_fps = {15, 24, 5, 30, 48, 50, 60, 44100, 48000}
+    float_fps = {23.976, 29.97, 29.97, 47.952, 59.94}
 
     fps = lib.get_project_fps()
 
-    if not isinstance(fps, (int, float)):
-        raise ValueError("Set value for project's FPS is not a number. "
-                         "Only accepts floats and integers")
+    if isinstance(fps, float) and fps in float_fps:
+        unit = "{:f}fps".format(int(fps))
```
BigRoy: This should not have the `int(fps)` :) otherwise the unit will not come out correctly, right?

colorbleed/maya/__init__.py
```diff
@@ -2,15 +2,14 @@
...
-from maya import cmds
+from maya import utils, cmds, mel
 
 from avalon import api as avalon, pipeline, maya
 from pyblish import api as pyblish
 
 from ..lib import (
     update_task_from_path,
-    any_outdated
+    any_outdated,
```
BigRoy: Redundant comma?

colorbleed/maya/lib.py
```diff
@@ -1310,3 +1311,76 @@ def _get_parent(node):
...
+        parent = top_level_widgets.get("MayaWindow", None)
+        if parent is None:
+            pass
+        else:
+            dialog = popup.Popup2(parent=parent)
+            dialog.setModal(True)
+            dialog.setWindowTitle("Maya scene not in line with project")
+            dialog.setMessage("The FPS is out of sync, please fix")
+
+            # Set new text for button (add optional argument for the popup?)
```
BigRoy: Comment is not accurate anymore

colorbleed/widgets/popup.py
```diff
@@ -104,6 +104,107 @@ def calculate_window_geometry(self):
...
+        parent = self.parent()
+        self.close()
+
+        # Trigger the signal
+        self.on_show.emit()
+
+        if parent:
+            parent.raise_()
+
+    def calculate_window_geometry(self):
```
BigRoy: Seem like code duplication? Does it make more sense to make a base Popup class that you can inherit from?

colorbleed/maya/lib.py
```diff
@@ -1310,3 +1311,76 @@ def _get_parent(node):
...
+    fps = lib.get_project_fps()
+
+    if isinstance(fps, float) and fps in float_fps:
+        unit = "{:f}fps".format(int(fps))
+
+    elif int(fps) in int_fps:
+        unit = "{:d}fps".format(int(fps))
+
+    else:
+        raise ("Unsupported FPS value: `%s`" % fps)
```
BigRoy: Always raise with a specified error type, e.g. `ValueError`!

colorbleed/plugins/maya/publish/validate_maya_units.py
```diff
@@ -44,7 +48,5 @@ def repair(cls):
         current_linear = cmds.currentUnit(query=True, linear=True)
         cls.log.debug(current_linear)
 
-        cls.log.info("Setting time unit to 'PAL'")
-        cmds.currentUnit(time="pal")
-        current_time = cmds.currentUnit(query=True, time=True)
-        cls.log.debug(current_time)
+        cls.log.info("Setting time unit to match project")
+        mayalib.set_project_fps()
```
BigRoy: How about letting the maya lib expose a more generic method `set_scene_fps(fps)` and you just pass the correct fps to the method? Would simplify it and make it more explicit on all ends?

colorbleed/maya/__init__.py
```diff
@@ -119,6 +126,9 @@ def on_open(_):
     from avalon.vendor.Qt import QtWidgets
     from ..widgets import popup
 
+    # Ensure scene's FPS is set to project config
+    lib.validate_fps()
```
BigRoy: If this is checked using `before_save_check` do you still want to do it `on_save` too? Seems redundant.

colorbleed/lib.py
```diff
@@ -245,3 +245,23 @@ def collect_container_metadata(container):
...
+    project = io.find_one({"name": project_name,
+                           "type": "project"},
+                          projection={"config": True})
+
+    config = project.get("config", None)
+    assert config, "This is a bug"
+
+    fps = config.get("fps", 25.0)
+
+    return float(fps)
```
BigRoy: Because you're forcing it to a float here it means that the `set_project_fps` will never actually trigger with and integer, not a problem - just wanted to state that. :)

colorbleed/maya/__init__.py
```diff
@@ -96,10 +97,16 @@ def _fn():
     safe_deferred(override_component_mask_commands)
 
 
+def before_save_check(return_code, _):
```
BigRoy: Shouldn't this match naming like the other methods, the `_check` seems redundant looking at the other methods like `on_save`, `on_load`, etc. :)

colorbleed/plugins/maya/publish/validate_maya_units.py
aardschok: Yes! Let's do it.

colorbleed/maya/__init__.py
aardschok: So you save a scene which has an invalid FPS set (which should possible as discussed yesterday), wouldn't you want to be reminded that the scene has been stored with an invalid FPS when opening? I can assume people (animators) will just work with it. This can result in timings which are wrong even when you update the keys to match the new animation which means extra work.

It is only displayed when the file being opened has been stored with the WRONG FPS. Yes it will be in your face until its fixed. This will be something that can be remedied with an override per shot if need be.

colorbleed/lib.py
aardschok: Might be we need to look at creating an FPS table. The following is occuring:
When your FPS is 25 in the scene, it displays as an integer. Using `mel.eval('currentTimeUnitToFPS()')` returns a float so the 25 FPS becomes 25.0. Maybe a float / int comparisonneeds to be added so we return whatever FPS has been set.

aardschok: The issue just got more annoying, running `currentTimeUnitToFPS` in mel it returns the correct values. When it's a float it returns a float, it its an int. Running the code in python through `mel.eval` always returns a float, It doesn't properly return the original value. Maybe you know why?

colorbleed/maya/lib.py
```diff
@@ -1354,9 +1353,20 @@ def validate_fps():
 
     """
 
+    fps = lib.get_project_fps()  # can be int or float
     current_fps = mel.eval('currentTimeUnitToFPS()')  # returns float
-    fps = lib.get_project_fps()
-    if fps != current_fps:
+
+    # Check if FPS is supported in Maya and compare with current fps
+    if isinstance(fps, int) and fps in INT_FPS:
```
BigRoy: The INT_FPS checks, etc. are redundant - should be removed.

It should just check whether `current_fps == project_fps`

```python
fps = lib.get_project_fps()  # can be int or float
current_fps = mel.eval('currentTimeUnitToFPS()')  # returns float

if current_fps != fps:
    # show warning
```

PR #126, opened 2018-07-02
Fixed loading Yeti nodes/cache

aardschok:
* Added check if `value` actually has a value (not None)
* Fixed `PLN-135`


BigRoy:
When is a value None on load? Doesn't that mean it's missing the cache??

aardschok:
When the `imageSearchPath` is not needed it won't be set and will return None when querying. It is not always the cache file specifically ;)

BigRoy:
I see, makes sense. :+1: Great.

PR #127, opened 2018-07-04
Extend validate render settings plugins

aardschok:
* Validate animation is toggled 
* Added repair option
  Does not repair the animation as there are multiple possibilties for it (standard vs specific range)

### DIFF
--- a/colorbleed/plugins/maya/publish/validate_rendersettings.py
+++ b/colorbleed/plugins/maya/publish/validate_rendersettings.py
@@ -16,8 +16,13 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
     * Frame Padding must be:
         * default: 4
 
+    * Animation must be toggle on, in Render Settings - Common tab:
+        * vray: Animation on standard of specific
+        * arnold: Frame / Animation ext: Any choice without "(Single Frame)"
+        * redshift: Animation toggled on
+
     NOTE:
-        The repair function of this plugin does not repair animation
+        The repair function of this plugin does not repair the animation
         setting of the render settings due to multiple possibilities.
 
     """
@@ -37,7 +42,7 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
         renderer = instance.data['renderer']
         layer_node = instance.data['setMembers']
 
-        # Collect the filename prefix in the renderlayer
+        # Collect the filename prefix in the render layer
         with lib.renderlayer(layer_node):
 
             render_attrs = lib.RENDER_ATTRS.get(renderer,
--- a/colorbleed/plugins/maya/publish/validate_rendersettings.py
+++ b/colorbleed/plugins/maya/publish/validate_rendersettings.py
@@ -17,7 +17,7 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
         * default: 4
 
     NOTE:
-        The repair function of this plugin will does not repair animation
+        The repair function of this plugin does not repair animation
         setting of the render settings due to multiple possibilities.
 
     """
@@ -37,11 +37,6 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
         renderer = instance.data['renderer']
         layer_node = instance.data['setMembers']
 
-        # Main check animation
-        animation = cmds.getAttr("defaultRenderGlobals.animation")
-        assert animation is True, ("Animation needs to be enabled in the "
-                                   "render settings")
-
         # Collect the filename prefix in the renderlayer
         with lib.renderlayer(layer_node):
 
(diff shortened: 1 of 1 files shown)


colorbleed/plugins/maya/publish/validate_rendersettings.py
```diff
@@ -31,6 +37,11 @@ def process(self, instance):
         renderer = instance.data['renderer']
         layer_node = instance.data['setMembers']
 
+        # Main check animation
+        animation = cmds.getAttr("defaultRenderGlobals.animation")
```
BigRoy: If you're checking per layer anyway - why does this need to also be checked globally? Just wondering.

colorbleed/plugins/maya/publish/validate_rendersettings.py
```diff
@@ -15,12 +15,18 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
 
     * Frame Padding must be:
         * default: 4
+
+    NOTE:
+        The repair function of this plugin will does not repair animation
```
BigRoy: What are these "multiple possibilities"?

Also, typo: ` will does not` -> ` will not`  or `does not`.

aardschok: > What are these "multiple possibilities"?

You have `Standard` and `Specific Frames` which both return `True` when querying the animation attribute of `defaultGlobals`. The difference lies in how to query the frame range. I have not dived in to this further.

colorbleed/plugins/maya/publish/validate_rendersettings.py
aardschok: When we change renderlayer the access to the `defaultRenderGlobals` is still valid and the renderlayer applies any overrides to the `defaultRenderGlobals` if it has any. So if someone, for some reason decides to make an override to the `animation` to `Disabled` (which is possible) we catch it as well.

BigRoy: Yes. I understand. But still, why then need to check it globally *before* switching to the renderlayer?

colorbleed/plugins/maya/publish/validate_rendersettings.py
BigRoy: Ah yes, for now we are not even supporting Specific Frames, correct?

Specific Frames can set e.g. 1, 3, 7, 10 (to just render those). The syntax also supports ranges. Pretty dope. Actually, thinking about it that's an issue on its own (for YouTrack?). Rarely, if ever, used by the renderteam. ;) Might also conflict with how we publish sequences currently - we only allow consecutive sequences.

colorbleed/plugins/maya/publish/validate_rendersettings.py
aardschok: After reconsidering my answer, you are correct. A bit weird as we also go over the `masterLayer`.

colorbleed/plugins/maya/publish/validate_rendersettings.py
aardschok: We will need to see how this effects the other renders. I noticed that Arnold for example uses the file name template from a dropdown menu to check if it can render animation. It needs to be set to `<fname>.#.<ext>` or something similar.

I do not know how Redshift goes about its way to recognize a sequence of frames. Will take a look at that now.

aardschok: Relieve! Redshift uses a simple checkbox for this which has the same name.

BigRoy:
If this is stable, let's give it a go in production.

colorbleed/plugins/maya/publish/validate_rendersettings.py
```diff
@@ -43,11 +49,40 @@ def process(self, instance):
...
-        assert prefix == fname_prefix, (
-            "Wrong file name prefix, expecting %s" % fname_prefix
-        )
-        assert padding == self.DEFAULT_PADDING, (
-            "Expecting padding of {} ( {} )".format(self.DEFAULT_PADDING,
-                                                    "0"*self.DEFAULT_PADDING)
-        )
+            anim_override = cmds.getAttr("defaultRenderGlobals.animation")
+            assert anim_override is True, (
+                "Animation neesd to be enabled. Use the same frame for start "
```
BigRoy: `neesd` -> `needs`

colorbleed/plugins/maya/publish/validate_rendersettings.py
```diff
@@ -43,11 +49,40 @@ def process(self, instance):
...
+                "and end to render singel frame")
```
BigRoy: `singel` -> `single`

PR #128, opened 2018-07-05
Add warning dialog for import

aardschok:
* Prevents the user from accidentally importing meshes
* Fixes issue 112


BigRoy:
I'm not a fan of this idea, but let's try it in production. I'm just afraid of someone importing 100 models and having to click to one hundred pop-ups. Feel free to merge if this is tested and works! :)

aardschok:
> I'm just afraid of someone importing 100 models 

Yes I see your point

PR #129, opened 2018-07-09
Fixed bug in attribute_values

aardschok:
The contextmanager function `attribute_values` had conflicting results when it came to resolving string type values. To resolve the issue I moved the function to `colorbleed.maya.lib` to be able to use `six.string_type`.

_Changes:_
* Added function to `colorbleed.maya.lib`
* Updated module who used that function

### DIFF
--- a/colorbleed/maya/lib.py
+++ b/colorbleed/maya/lib.py
@@ -13,6 +13,7 @@ from collections import OrderedDict, defaultdict
 from maya import cmds, mel
 
 from avalon import api, maya, io, pipeline
+from avalon.vendor.six import string_types
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


BigRoy:
What was causing this? These extractors would only ever run in Python 2 (because Maya?) so should still never happen.

Do you have an example case?

aardschok:
[Screenshot](https://gyazo.com/85ff22d56e267a3070859ae5186500b0)

Code case:
```
       # Here we explicitly force str to rule out any unicode issues
        attr_value = {"%s.imageSearchPath" % n: str(image_search_path) for
                            n in yeti_nodes}

        # Get input_SET members
        # Get all items
        set_members = cmds.sets(input_set[0], query=True)
        members = cmds.listRelatives(set_members, ad=True, fullPath=True) or []
        members += cmds.ls(set_members, long=True)

        nodes = instance.data["setMembers"]
        with disconnected_attributes(settings, members):
            with maya.attribute_values(attr_value):  # issue is here where we apply any changes
```

BigRoy:
Shouldn't unicode then just be captured as forced to str in the attribute values. Not sure why it would six if it's only for compatibility between Python 2 and 3. :) Anyway, fine if it works - just seems like an odd dependency for purely a Python 2 codebase (Maya)

PR #130, opened 2018-07-10
Fix bug in update

aardschok:
When updating an asset through the manager we check the reference nodes. The check for `sharedReferenceNode` was hardcoded.

* Updated check for `sharedReferenceNode`
* Minor cosmetics
* Updated docstring


BigRoy:
Shouldn't the startswith be "sharedReferenceNode"?

PR #131, opened 2018-07-12
Implement Extend Frames 

aardschok:
Allow the user to render new frames to extend the shot either at the beginning or end of the sequence.
This is done by collecting the render layer and it's render passes (`Render Elements` or `AOVs`), check their previous version and collect the files from that version. We use `clique` to create the list of files to copy prior to rendering the new frames. The copied frames and newly rendered frames create a new version together which is then published as a whole.

__Updated:__
Collect render layer AOVs
  * Support for VRay / Arnold / Redshift

Validate render layer AOVs
  * Check if result of `<layer>.<aov>` is registered, logic won't work otherwise

Submit publish job
  * get version per subset (renderlayer , renderpass)
  * get files per subset
  * copy found files

Render Globals
  * Added attributes to support extend logic

__Dependencies:__
[Maya Deadline Submission tool's UI](https://github.com/Colorbleed/maya-deadline-submission-settings/pull/7)

### DIFF
--- a/colorbleed/plugins/global/publish/submit_publish_job.py
+++ b/colorbleed/plugins/global/publish/submit_publish_job.py
@@ -1,5 +1,6 @@
 import os
 import json
+import pprint
 import re
 
 from avalon import api, io
@@ -198,7 +199,7 @@ class SubmitDependentImageSequenceJobDeadline(pyblish.api.InstancePlugin):
             # Frame comparison
             prev_start = None
             prev_end = None
-            resource_range = range(int(start)-1, int(end)+1)
+            resource_range = range(int(start), int(end)+1)
 
             # Gather all the subset files (one subset per render pass!)
             subset_names = [data["subset"]]
@@ -215,16 +216,36 @@ class SubmitDependentImageSequenceJobDeadline(pyblish.api.InstancePlugin):
                     prev_end = version["data"]["endFrame"]
 
                 subset_resources = get_resources(version, _ext)
-                get_resource_files(subset_resources, resource_range, override)
+                resource_files = get_resource_files(subset_resources,
+                                                    resource_range,
+                                                    override)
 
-                resources.extend(subset_resources)
+                resources.extend(resource_files)
 
-            updated_start = start if start < prev_start else prev_start
-            updated_end = end if end > prev_end else prev_end
+            updated_start = min(start, prev_start)
+            updated_end = max(end, prev_end)
 
             # Update metadata and instance start / end frame
             self.log.info("Updating start / end frame : "
                           "{} - {}".format(updated_start, updated_end))
+
+            # TODO : Improve logic to get new frame range for the
+            # publish job (publish_filesequence.py)
+            # The current approach is not following Pyblish logic which is based
+            # on Collect / Validate / Extract.
+
+            # ---- Collect Plugins  ---
+            # Collect Extend Frames - Only run if extendFrames is toggled
+            # # # Store in instance:
+            # # # Previous rendered files per subset based on frames
+            # # # --> Add to instance.data[resources]
+            # # # Update publish frame range
+
+            # ---- Validate Plugins ---
+            # Validate Extend Frames
+            # # # Check if instance has the requirements to extend frames
+            # There might have been some things which can be added to the list
+            # Please do so when fixing this.
 
             # Start frame
             metadata["startFrame"] = updated_start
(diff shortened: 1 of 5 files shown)


colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -114,6 +182,58 @@ def process(self, instance):
...
+                                             subset_name=subset_name,
+                                             family=family)
+
+                # Set prev start / end frames for comparison
+                if not prev_start and not prev_end:
+                    prev_start = version["data"]["startFrame"]
+                    prev_end = version["data"]["endFrame"]
+
+                subset_resources = get_resources(version, _ext)
+                get_resource_files(subset_resources, resource_range, override)
```
BigRoy: The return value is not captured - is this function redundant?

colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -114,6 +182,58 @@ def process(self, instance):
...
+
+                resources.extend(subset_resources)
+
+            updated_start = start if start < prev_start else prev_start
```
BigRoy: This can be just `min(start, prev_start)`

colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -114,6 +182,58 @@ def process(self, instance):
...
+            updated_end = end if end > prev_end else prev_end
```
BigRoy: This can be just `max(end, prev_end)`

colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -159,3 +279,22 @@ def process(self, instance):
         response = requests.post(url, json=payload)
         if not response.ok:
             raise Exception(response.text)
+
+        # Copy files from previous render if extendFrame is True
+        self.log.info("Submitting done!")
+
+        # return
```
BigRoy: Redundant comment line

colorbleed/plugins/maya/publish/validate_renderlayer_aovs.py
```diff
@@ -0,0 +1,60 @@
...
+    Each render element is registered as a subset which is formatted based on
+    the render layer and the render element, example:
+
+        <render layer>.<render element>
+
+    This translates to something like this:
+
+        CHAR.diffuse
+
+    This check is only used when the Extend Frames in 'renderGlobalsDefault'
```
BigRoy: Is this true, I don't see this check being done within this validator?

colorbleed/plugins/maya/publish/validate_renderlayer_aovs.py
```diff
@@ -0,0 +1,60 @@
...
+    def process(self, instance):
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise RuntimeError("Found unregistered subsets: {}".format(invalid))
+
+    def get_invalid(self, instance):
+
+        invalid = []
+
+        asset_name = api.Session["AVALON_ASSET"]
```
BigRoy: This is actually incorrect - the asset that will be published should be present inside the instance upon Collection - as such it should actually use the `instance.data["asset"]` or whatever the key is called, right?

colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -114,6 +182,58 @@ def process(self, instance):
...
+
+            updated_start = start if start < prev_start else prev_start
+            updated_end = end if end > prev_end else prev_end
+
+            # Update metadata and instance start / end frame
+            self.log.info("Updating start / end frame : "
+                          "{} - {}".format(updated_start, updated_end))
+
+            # Start frame
+            metadata["startFrame"] = updated_start
```
BigRoy: This overriding of the start and end frame seems right - but it feels off that we're doing this all the way in the Integrator. Logically this should actually be at the point of "collecting" since it's part of knowing how it should be published. But I can see how this is an edge case scenario and is alright to behave a bit oddly - it just feels so weird at this point in time.

Just wanted to have it noted.

aardschok: Yes I agree and it felt weird to place it here but the situation is a bit more complex than just checking ranges.

We need to have the subset's version to retrieve the frame range. 
This means validating if the render pass is an actual subset will need to be prior to the collecting of the render.

The logic is a bit intertwined, I am open to suggestions to improve this. Maybe allow a mediator plugin to handle both the validation and collection.

PR #139, opened 2018-07-24
Fixed updating Yeti cache

aardschok:
Extended the update logic for the Yeti Cache loader. It was not properly adding or removing the nodes of the container(s).

It is based on the content of the container and the metadata which is published along with the cache. Both the container and the metadata are used to create a lookup with `defaultdict` based on `cbId`. This way we support renaming and adjusting the correct nodes. All IDs which are not in the metadata look up are to be deleted, all new IDs in the metadata file are created and added to the container.

### DIFF
--- a/colorbleed/plugins/maya/load/load_yeti_cache.py
+++ b/colorbleed/plugins/maya/load/load_yeti_cache.py
@@ -121,7 +121,8 @@ class YetiCacheLoader(api.Loader):
                 # Get all related nodes
                 shapes = scene_lookup[_id]
                 # Get the parents of all shapes under the ID
-                transforms = cmds.listRelatives(shapes, parent=True,
+                transforms = cmds.listRelatives(shapes,
+                                                parent=True,
                                                 fullPath=True) or []
                 to_remove.extend(shapes + transforms)
 
@@ -138,6 +139,9 @@ class YetiCacheLoader(api.Loader):
             data["attrs"]["cacheFileName"] = os.path.join(path, cache_file_path)
 
             if cb_id not in scene_lookup:
+
+                self.log.info("Creating new nodes ..")
+
                 new_nodes = self.create_nodes(namespace, [data])
                 cmds.sets(new_nodes, addElement=container_node)
                 cmds.parent(new_nodes, container_root)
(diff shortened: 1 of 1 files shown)


colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -83,30 +84,88 @@ def remove(self, container):
...
+
+                for scene_node in scene_nodes:
+                    # Remove the namespace root ":namespace:"
+                    compare_name = scene_node.split(namespace)[-1][1:]
+                    if compare_name != match_name:
+                        self.log.info("Renaming %s to %s" %
+                                      (scene_node, match_name))
+
+                        # Get transform node, this makes renaming easier
+                        transform_node = scene_node.rstrip("Shape")
```
BigRoy: Let's do `maya.cmds.listRelatives` similar to [here](https://github.com/Colorbleed/colorbleed-config/pull/139/files#diff-e54e80d1c1b4885422ca99bcc6b69d44R124)

colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -83,30 +84,88 @@ def remove(self, container):
...
+        for cb_id, data in meta_data_lookup.items():
+
+            # Update cache file name
+            file_name = data["name"].replace(":", "_")
+            cache_file_path = "{}.%04d.fur".format(file_name)
+            data["attrs"]["cacheFileName"] = os.path.join(path, cache_file_path)
+
+            if cb_id not in scene_lookup:
+                new_nodes = self.create_nodes(namespace, [data])
+                cmds.sets(*new_nodes, addElement=container_node)
```
BigRoy: Doesn't this also work without the `*`? Like `cmds.sets(new_nodes, addElement=container_node)`

colorbleed/plugins/maya/load/load_yeti_cache.py
```diff
@@ -152,70 +219,56 @@ def validate_cache(self, filename, pattern="%04d"):
...
+                cache = os.path.join(self.fname,
+                                     "{}.%04d.fur".format(file_name))
+                self.validate_cache(cache)
+                attributes["cacheFileName"] = cache
+
+            # Update attributes with requirements
+            attributes.update({"viewportDensity": 0.1,
+                               "verbosity": 2,
+                               "fileMode": 1,
+                               "visibleInReflections": True,
```
BigRoy: Nice to see this fixed - but I think it is actually more readable to keep these two "fixes" separate from the attributes update.

How about we do it explicitly like how Sven did it in his PR.

```python
            # Fix for : YETI-6
            # Fixes the render stats (this is literally taken from Perigrene's
            # ../scripts/pgYetiNode.mel script)
            cmds.setAttr(yeti_node + ".visibleInReflections", 1)
            cmds.setAttr(yeti_node + ".visibleInRefractions", 1)
```

colorbleed/plugins/maya/load/load_yeti_cache.py
aardschok: Yes it should, don't know why I used this one :thinking:

PR #143, opened 2018-08-13
Show Workfiles on launch.

tokejepsen:
This PR is just a suggestion, so feel free to close it.

I figured Colorbleed might be faster at testing this app, than I would since you have more people running Avalon on a daily basis.

This does require an update of the core.


tokejepsen:
No problem. Happy to see it being useful.

tokejepsen:
Closing this PR as it needs conflict resolution and I'm not using colorbleed-config. The code is there for reference.

PR #145, opened 2018-08-22
Fix Yeti Cache update #2

BigRoy:
And another implementation of the Yeti cache fixes by Sven, hehe. Let's choose.


BigRoy:
Thanks @svenneve. 

We'll be continuing with @aardschok's version but steal some bits from your code. :) So we'll close this PR. 

However it would still be good to have the `menu.json` changes - can we set up a separate PR for that?

PR #147, opened 2018-08-24
Skip unconnected / invalid renderlayer

aardschok:
Resolved issue PLN-146

In order to combat the issue of non-accessible renderlayers we have extended the lookup for renderlayers.
Prior to retrieving all renderlayers we check which renderlayers are connected to the renderlayer manager.
Before collecting all information we check if the renderable renderlayers are connected to the renderlayer manager. Any renderlayers which are not connected will be skipped with a warning message which is visible in the Pyblish window.

Example of a skipped renderlayer
[Image from Gyazo](https://gyazo.com/d58017f33d0ce3546dd2b128e0a81c71)


colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -9,7 +9,7 @@
 class CollectMayaRenderlayers(pyblish.api.ContextPlugin):
     """Gather instances by active render layers"""
 
-    order = pyblish.api.CollectorOrder
+    order = pyblish.api.CollectorOrder + 0.01
```
BigRoy: Why in particular was this change required?

aardschok: Ah, this is remnant of the previous approach where I first checked if renderlayers were accessible. I will remove it

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -32,7 +32,12 @@ def process(self, context):
         context.data["startFrame"] = start_frame
         context.data["endFrame"] = end_frame
 
-        # Get render layers
+        # Get all linked renderlayers
```
BigRoy: Shall we add another comment line here?
```python
    # Get all valid renderlayers
    # This is how Maya populates the renderlayer display
```

Since it's such a weird bit of code I want to emphasize that its based on some code from how Maya does it.

BigRoy:
<3 Nice work @aardschok

PR #148, opened 2018-08-24
Expose primary and secondary pool

aardschok:
In order to reduce the artists time look up current submission settings we added the primary and secondary pool to the `renderglobalsDefault`. Added backwards compatibility.

This PR is in conjunction with : [Maya Deadline Submission tool](https://github.com/Colorbleed/maya-deadline-submission-settings/pull/11)


colorbleed/plugins/global/load/copy_file.py
```diff
@@ -32,4 +32,4 @@ def copy_file_to_clipboard(path):
 
         # Set to Clipboard
         clipboard = app.clipboard()
-        clipboard.setMimeData(mime)
+        clipboard.setMimeData(mime)
```
BigRoy: Blegh, is this pycharm doing that? :) Shouldn't it actually force a new line being there?

colorbleed/__init__.py
```diff
@@ -24,3 +24,8 @@ def uninstall():
     print("Deregistering global plug-ins..")
     pyblish.deregister_plugin_path(PUBLISH_PATH)
     avalon.deregister_plugin_path(avalon.Loader, LOAD_PATH)
+
+
+def register_loader_plugins():
```
BigRoy: Is this needed for this particular PR? Seems like a remainder from something else?

colorbleed/plugins/maya/publish/collect_renderlayers.py
```diff
@@ -154,7 +152,33 @@ def parse_options(self, render_globals):
...
+        # Check for specific pools
+        if "primaryPool" in attributes:
+            pool_a = attributes["primaryPool"]
+            pool_b = attributes["secondaryPool"]
+
+        else:
+            # Backwards compatibility
+            pool_str = attributes.get("pools", None)
+            if pool_str:
+                print("XXX", pool_str.split(";"))
```
BigRoy: Debug print statement should be removed?

BigRoy:
Just some minor comments, but functionality seems to be good outside of those. Nice!

colorbleed/__init__.py
aardschok: This is a remainder of the standalone plugins which we have discussed in PLN-105

PR #150, opened 2018-08-27
Submit .vrscene to Deadline

aardschok:
This PR introduces the ability to submit a Maya scene to be exported to `.vrscene`. This will be done by creating a new instance called `VRay Scene`.

The VRay Scene instance gathers the following information:
* Render Layers
* Start and end frame from render settings
* Resolution (from `defaultResolution` node)

It is checked for the following things:
* A `VRaySettings` node
* Export to `.vrscene` is toggled on (attribute `vrscene_on`)
* Each Frame in File is enabled (attribute `misc_eachFrameInFile`)
* The template for export is correct (attribute `vrayscene_filename`)
  This should be: `vrayscene/<Scene>/<Scene>_<Layer>/<Layer>`

The Deadline job created is a 3 step process:
_The process is repeated for each renderlayer_
1. Export a `.vrscene` with files per frame
  This task is limited to one machine
2. VRay Render job (dependend on 1)
  The tasks (frames) are distributed over multiple machines.  
3. Publish Job (depended on 2 and `metadata.json`)
  A single task to publish the frames rendered based on the metadata


colorbleed/plugins/maya/create/colorbleed_vrayscene.py
```diff
@@ -0,0 +1,38 @@
...
+        # We won't be publishing this one
+        self.data["id"] = "avalon.vrayscene"
+
+        # We don't need subset or asset attributes
+        self.data.pop("subset", None)
+        self.data.pop("asset", None)
+
+        data = OrderedDict(**self.data)
+
+        data["camera"] = self._get_cameras()
```
BigRoy: @aardschok does this work? This means it would just end up become a formatted list in the string like `["cam1", "cam2"]` over [here](https://github.com/Colorbleed/colorbleed-config/pull/150/files#diff-abe4cfcd94241bfaed3e0e5865835bb0R195)

colorbleed/plugins/global/publish/submit_publish_job.py
```diff
@@ -134,8 +136,10 @@ def process(self, instance):
         # Get a submission job
         job = instance.data.get("deadlineSubmissionJob")
         if not job:
-            raise RuntimeError("Can't continue without valid deadline "
-                               "submission prior to this plug-in.")
+            self.log.warning("Can't continue without valid deadline "
```
BigRoy: @aardschok Why shouldn't this be a raised error? :) Since it is actually failing completely.

colorbleed/plugins/maya/create/colorbleed_vrayscene.py
aardschok: No. The `lib.read` function returns the string of the index which has been set.

This example has two renderable cameras added.
```
lib.read("vraysceneDefault")
# Result: {u'suspendPublishJob': False, u'suspendRenderJob': False, u'family': u'colorbleed.vrayscene', u'pools': u'', u'extendFrames': False, u'camera': u'perspShape', u'active': True, u'id': u'avalon.vrayscene'} # 
setAttr "vraysceneDefault.camera" 1;
lib.read("vraysceneDefault")
# Result: {u'suspendPublishJob': False, u'suspendRenderJob': False, u'family': u'colorbleed.vrayscene', u'pools': u'', u'extendFrames': False, u'camera': u'perspShape2', u'active': True, u'id': u'avalon.vrayscene'} # 
```

PR #152, opened 2018-09-06
Add control for samples, Yeti

aardschok:
Resolves issue YETI-7

Adds the option for the user to set the amount of sample for the Yeti cache.
If the attribute value is zero it will fallback to `sampleTimes : 0.0 1.0`, its default value is 3 which is similar to the default settings of the pgYetiMaya node,

* Added attribute to the instance
* Added support and backwards compatibility

### DIFF
--- a/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
+++ b/colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
@@ -15,11 +15,12 @@ class CreateYetiCache(avalon.maya.Creator):
     def __init__(self, *args, **kwargs):
         super(CreateYetiCache, self).__init__(*args, **kwargs)
 
-        data = OrderedDict(self.data)
+        data = OrderedDict(**self.data)
         data["peroll"] = 0
 
         anim_data = lib.collect_animation_data()
         data.update({"startFrame": anim_data["startFrame"],
-                     "endFrame": anim_data["endFrame"]})
+                     "endFrame": anim_data["endFrame"],
+                     "samples": 3})
 
         self.data = data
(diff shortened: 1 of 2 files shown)


colorbleed/plugins/maya/create/colorbleed_yeti_cache.py
```diff
@@ -15,11 +15,12 @@ class CreateYetiCache(avalon.maya.Creator):
     def __init__(self, *args, **kwargs):
         super(CreateYetiCache, self).__init__(*args, **kwargs)
 
-        data = OrderedDict(self.data)
+        data = OrderedDict(**self.data)
```
BigRoy: @aardschok Why did this need to change? Didn't the previous work or what happened? :)

colorbleed/plugins/maya/publish/extract_yeti_cache.py
```diff
@@ -37,16 +37,23 @@ def process(self, instance):
         if preroll > 0:
             start_frame -= preroll
 
+        kwargs = {}
+        samples = instance.data.get("samples", 0)
+        if samples == 0:
+            kwargs.update({"sampleTimes": "0.0 1.0"})
```
BigRoy: If this is redundant (if `"sampleTimes": "0.0 1.0"` is the same as `"samples": 0`) then this kwargs could be removed and just `samples` can be passed to the `pgYetiCommand` - less code and more readable. :)

BigRoy: @aardschok did you ever figure out whether [this change](https://github.com/Colorbleed/colorbleed-config/pull/152#discussion_r215656795) would actually behave the same? If so let's simplify this code and avoid the _if..else_ statement.

PR #154, opened 2018-09-06
Add draft for labeling asset and frame range in Pyblish instance name

BigRoy:
New instance labels see:
image

### DIFF
--- a/colorbleed/plugins/maya/publish/collect_instances.py
+++ b/colorbleed/plugins/maya/publish/collect_instances.py
@@ -102,9 +102,18 @@ class CollectInstances(pyblish.api.ContextPlugin):
             parents = self.get_all_parents(members)
             members_hierarchy = list(set(members + children + parents))
 
-            # Create the instance
+            # Define nice label
             name = cmds.ls(objset, long=False)[0]   # use short name
-            instance = context.create_instance(data.get("name", name))
+            label = "{0} ({1})".format(name,
+                                       data["asset"])
+
+            # Append start frame and end frame to label if present
+            if "startFrame" and "endFrame" in data:
+                label += "  [{0}-{1}]".format(int(data["startFrame"]),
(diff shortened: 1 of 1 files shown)


BigRoy:
Not sure if code could be nicer. Ideas?

colorbleed/plugins/maya/publish/collect_instances.py
```diff
@@ -102,9 +102,18 @@ def process(self, context):
...
+            # Define nice label
             name = cmds.ls(objset, long=False)[0]   # use short name
-            instance = context.create_instance(data.get("name", name))
+            label = "{0} ({1})".format(name,
+                                       data["asset"])
+
+            # Append start frame and end frame to label if present
+            if "startFrame" and "endFrame" in data:
+                label += "  [{0}-{1}]".format(int(data["startFrame"]),
+                                              int(data["endFrame"]))
```
aardschok: I have a question regarding start and end frames, when does it occur that the frame numbers need to be floats?
If we could counter that this solution would be nicer: `" [{startFrame:d}-{endFrame:d}]".format(**data)`

No further objections :+1:

BigRoy: They are floats, because they could support substeps (technically) all-round on the pipeline. Never used it really, but technically there's nothing stopping us.

Unfortunately I was unable to find a formatting version that would force integers that worked both with floats and integers. Your above example fails on floats for example, like:

```python
>>> print "{0:d}".format(0.0)

Traceback (most recent call last):
  File "<pyshell#0>", line 1, in <module>
    print "{0:d}".format(0.0)
ValueError: Unknown format code 'd' for object of type 'float'
```

So forcing to `int()` was the only way that made sense. :) Agree?

BigRoy: Thanks, merging this to give it a go in production.

PR #157, opened 2018-09-10
Add on_new callback for Maya

aardschok:
Resolves issue PLN-159

__Note: This requires the projects to be reconfigured. fps will need to move to `project.data`__

In order to ensure that the user's project settings are applied after launching Maya we added the `on_new` callback which will apply the project's settings such as FPS and resolution.

This PR also introduces a minor refactoring in the way the project information is being queried. By adding the `get_project_data` we reduce the number of specific calls which eventually are stored in the same `data`.

The `get_project_fps` will call `get_project_data` first and then retrieve the FPS out of it. The idea is to get and set all other relevant settings when launching Maya.

### DIFF
--- a/colorbleed/lib.py
+++ b/colorbleed/lib.py
@@ -257,16 +257,33 @@ def get_project_fps():
 
     Returns:
         int, float
+
+    """
+
+    data = get_project_data()
+    fps = data.get("fps", 25.0)
+
+    return fps
+
+
+def get_project_data():
+    """Get the data of the current project
+
+    The data of the project can contain things like:
+        resolution
+        fps
+        renderer
+
+    Returns:
+        dict:
+
     """
 
     project_name = io.active_project()
     project = io.find_one({"name": project_name,
                            "type": "project"},
-                          projection={"config": True})
+                          projection={"data": True})
 
-    config = project.get("config", None)
-    assert config, "This is a bug"
+    data = project.get("data", {})
 
-    fps = config.get("fps", 25.0)
-
-    return fps
+    return data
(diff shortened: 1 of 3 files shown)


colorbleed/maya/lib.py
```diff
@@ -1384,7 +1385,8 @@ def set_scene_fps(fps, update=True):
     """
 
     if fps in FLOAT_FPS:
-        unit = "{:f}fps".format(fps)
+        # :.2f means <any digits>.<round to 2 digits>
+        unit = "{:.2f}fps".format(fps)
```
BigRoy: Does this still work for `23.976` and `47.952` since they don't match this amount of decimal points as described here?

I think if we remove the forced float formatting this should actually work, so just `"{}fps".format(fps)`

colorbleed/lib.py
```diff
@@ -257,16 +257,66 @@ def get_project_fps():
...
+    asset_name = asset or avalon.api.Session["AVALON_ASSET"]
+    document = io.find_one({"name": asset_name,
+                            "type": "asset"})
+
+    data = document.get("data", {})
+
+    return data
+
+
+def get_asset_fps():
```
BigRoy: I think this method is now redundant with [this line](https://github.com/Colorbleed/colorbleed-config/pull/157/files#diff-b9b06b643a808ffdecb81fd665eec054R1454) so let's remove this function.

BigRoy:
Nice - this is great! Almost ready for a merge, thanks! :)

colorbleed/lib.py
aardschok: This would also count for project fps, right?

BigRoy: Now that you mention it - I suppose so?

BigRoy: You're correct I think. It should also "behave" correctly for the "on_open", "on_save" that it knows to grab it from asset first, then project.

Well spotted.

PR #159, opened 2018-09-11
Also ignore _UNKNOWN_REF_NODE_1, or other variations

BigRoy:
Also capture _UNKNOWN_REF_NODE_1, or other variations (noticed that being present in one scene)


BigRoy:
Little fix on my last code.

PR #160, opened 2018-09-12
Introduce Houdini

aardschok:
This PR adds Houdini to the available hosts

Available after merge:
* Avalon menu for Houdini (xml file which needs to be copied)
* Basic plugins
  * Create Pointcache
  * Load Alembic
  * Validators for point cache
  * Extract Point cache


BigRoy:
Great stuff. I can see this is a draft implementation, still very nice. 

I'll try to get comments on this soon since I saw some things that felt out of the ordinary or dangerous down the line.

First, though, it sometimes still mentions "Maya" on class names or logged messages. Of course that is based on duplicated code for now, just wanted to give you heads up that is in there - especially the clashing class name gould give you a headache at some point if the plug-ins somehow get registered at the same time.

:+1:

PR #161, opened 2018-09-13
Add renderable camera checks to render pipeline in Maya

BigRoy:


BigRoy:
Tested here - merging now.

PR #162, opened 2018-09-13
Minor fixes for FPS + renderlayers + renderlayer nice label names

BigRoy:
Some fixes for the pipeline:

- Validate FPS on scene open after updating the task from the new scene so it tries to validate with the correct asset.
- FPS publish check to also use asset fps settings from database
- Fix getting AOVs code for Maya+V-Ray (support light select set node and non-explicit named extratex)
- Add nice label name for collecting renderlayers in Pyblish+Maya (with asset name and frame name)

### DIFF
--- a/colorbleed/plugins/maya/publish/collect_render_layer_aovs.py
+++ b/colorbleed/plugins/maya/publish/collect_render_layer_aovs.py
@@ -41,7 +41,7 @@ class CollectRenderLayerAOVS(pyblish.api.InstancePlugin):
 
         self.log.info("Renderer found: {}".format(renderer))
 
-        rp_node_types = {"vray": "VRayRenderElement",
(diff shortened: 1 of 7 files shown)


colorbleed/plugins/maya/publish/collect_render_layer_aovs.py
```diff
@@ -41,7 +41,7 @@ def process(self, instance):
 
         self.log.info("Renderer found: {}".format(renderer))
 
-        rp_node_types = {"vray": "VRayRenderElement",
+        rp_node_types = {"vray": ("VRayRenderElement", "VRayRenderElementSet"),
```
aardschok: For code consistency it would be nice to assume all values in this dictionary are lists or tuples. This would also indicate to the next person that it support multiple nodes, which the `type` flag of the `cmds.ls()` supports

BigRoy:
Thanks, done!

PR #163, opened 2018-09-18
Improvements on Houdini pipeline

aardschok:
Creator plugins for Houdini:
* VDB Cache 

Loader plugins:
* Load Camera (abc) for Houdini
* Load Alembic for Houdini:
  * added `colorbleed.model` family
* Load VDB Cache (vdb) for Maya VRay

Publish plugins:
* Collectors:
  * Animation data for all instances
  * Output of render node

* Validators:
  * Alembic output

* Extractor for VDB (Houdini)


tokejepsen:
You are probably developing to the needs of the production, but here are some plugin ive used at different companies: https://github.com/bumpybox/pyblish-bumpybox

PR #171, opened 2018-09-27
Optimize get_id (+/- 3x faster) and set_id (minor tweak)

BigRoy:
Optimization for a function that's run very often. :)

### DIFF
--- a/colorbleed/maya/lib.py
+++ b/colorbleed/maya/lib.py
@@ -11,6 +11,7 @@ import contextlib
 from collections import OrderedDict, defaultdict
 
 from maya import cmds, mel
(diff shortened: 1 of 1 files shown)


colorbleed/maya/lib.py
```diff
@@ -837,6 +847,7 @@ def set_id(node, unique_id, overwrite=False):
 
     # Set the value
     if not exists or overwrite:
+        attr = "{0}.cbId".format(node)
```
aardschok: If we go for performance we could go for the old skool formatting approach, `%s.cbId % node` . This is also micro improvements ;)

PR #172, opened 2018-09-28
Add support for vertex colors

aardschok:
Resolves issue MOD-09

* Include `writeColorSets` argument for Alembic export in instance
* Added control for user to disable `writeColorSets`

### DIFF
--- a/colorbleed/plugins/maya/create/colorbleed_model.py
+++ b/colorbleed/plugins/maya/create/colorbleed_model.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 import avalon.maya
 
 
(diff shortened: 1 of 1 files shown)


colorbleed/plugins/maya/create/colorbleed_model.py
```diff
@@ -8,3 +10,14 @@ class CreateModel(avalon.maya.Creator):
     label = "Model"
     family = "colorbleed.model"
     icon = "cube"
+
+    def __init__(self, *args, **kwargs):
+        super(CreateModel, self).__init__(*args, **kwargs)
+
+        # create an ordered dict with the existing data first
+        data = OrderedDict(**self.data)
```
BigRoy: @aardschok I think with [this change](https://github.com/getavalon/core/commit/3ecadff13427ff8239bca73168676902d028097c) in core, merged with [this PR](https://github.com/getavalon/core/pull/326) we wouldn't need to put this hack in code there, since it is an OrderedDict by default. :)

Haven't tested it though.

BigRoy:
:heart: Thanks!

colorbleed/plugins/maya/create/colorbleed_model.py
aardschok: Completely forgot about this :) thanks for pointing towards it. I will update some creator plugins :)

PR #179, opened 2018-10-05
Fix instance label (avoid changing the 'name' of the instance) and fix v-ray proxy validator

BigRoy:
- This reverts the `instance.name` or `instance.data["name"]` being changed to the nice label, instead it uses the `instance.data["label"]` that was intended for that.
- This fixes the V-Ray Proxy validator
- This removes the frame range validation from the V-Ray Proxy validator, but the frame range validator does it instead.


BigRoy:
With https://github.com/Colorbleed/colorbleed-config/pull/179/commits/b97b784255199baf15344b3509ec57ac652d4c15 the Generate UUIDs action on failed instances (for node ids) will now ignore any referenced nodes. As such it will never regenerate on any referenced node.

This also fixes LKD-23.

PR #184, opened 2018-10-10
REN-55: Avoid scene open/save/new callbacks during Maya batch rendering.

BigRoy:
Draft setup to fix REN-55.

_untested still though!_


BigRoy:
First tests seem positive, merging this.

PR #185, opened 2018-10-11
Add render callback validator

aardschok:
Resolves internal issue PLN-120

In order to ensure the render tasks are executed properly we need to check if the pre and post render callbacks are actually used.

For example:
Yeti is not loaded but its callback scripts are still set in the render settings. This will cause an error because Maya tries to find and execute the callbacks.


BigRoy:
Nice work @aardschok!

aardschok:
That is possible, I tried to make a more global plugin but I that would mean a larger look up table to manage.

PR #187, opened 2018-10-11
Set primitve to detail pattern at creation

aardschok:
Resolves internal issue PLN-176

At instance creation it will now set `prim_to_detail_pattern` to `cbId` .
This is needed to maintain the `cbId` after geometry is extracted.


aardschok:
I think there is an option to lock the attribute, that would be ideal

PR #192, opened 2018-10-16
Add attribute control for point cache

aardschok:
Fix internal issue: PLN-80

This PR will give the user extra control on which attribute(s) are exported along when writing a point cache. This will ensure that control attributes for shaders or FX are consistent during production.

* Improved creator plugin
* Fetch controls from instance for export

### DIFF
--- a/colorbleed/plugins/maya/create/colorbleed_pointcache.py
+++ b/colorbleed/plugins/maya/create/colorbleed_pointcache.py
@@ -1,5 +1,3 @@
-from collections import OrderedDict
-
 import avalon.maya
 from colorbleed.maya import lib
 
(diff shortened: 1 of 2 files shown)


colorbleed/plugins/maya/publish/extract_pointcache.py
```diff
@@ -32,6 +32,12 @@ def process(self, instance):
             start -= handles
             end += handles
 
+        attrs = instance.data.get("attr", "").split(";")
+        if not attrs:
+            attrs = ["cbId"]
```
BigRoy: Actually, `cbId` should *always* be included. As such even when `attr` is specified the `cbId` should be appended to the list. I guess it comes down to always doing `attrs += ["cbId"]`

colorbleed/plugins/maya/create/colorbleed_pointcache.py
```diff
@@ -15,22 +13,14 @@ class CreatePointCache(avalon.maya.Creator):
     def __init__(self, *args, **kwargs):
         super(CreatePointCache, self).__init__(*args, **kwargs)
 
-        # create an ordered dict with the existing data first
-        data = OrderedDict(**self.data)
+        data = {"writeColorSets": False,  # Vertex colors with the geometry.
+                "renderableOnly": False,  # Only renderable visible shapes
+                "visibleOnly": False,  # only nodes that are visible
+                "attr": "cbId",  # Add options for custom attributes
```
BigRoy: Since `cbId` is crucial to the pipeline I would force it in [here](https://github.com/Colorbleed/colorbleed-config/pull/192/files#r225471259) and remove it from needing to be included in the `attr` data on the instance. :)

What do you think @aardschok ?

BigRoy:
Nice stuff. Can we also get this added as support to the `model` family? (for the Alembic generated from there) or don't we need that?

aardschok:
> support to the model family?

This is an easy adjustment and makes things more streamlined and predictable. I say yes.

BigRoy:
:+1: Nice!

> This is an easy adjustment and makes things more streamlined and predictable. I say yes.

Let's add it to `CreateModel` too then. :)

PR #193, opened 2018-10-18
Patch for breaking validator

aardschok:
Updated the logic to validate the VRay translator settings. 

1. There is bug in pyblish ( fix is ready [here](https://github.com/pyblish/pyblish-base/pull/325) and [here](https://github.com/pyblish/pyblish-qml/pull/266) ) which allows all context plugins to ignore the `family`. Because of this the `validate_vray_translator_settings` plugin was being triggered even when having no related instance / family.

2. There was an issue with the output name from the VRay Standalone renderer. The output file name for the JobInfo previously `<layername>.%04d.<ext>` but the output generated was `<layername>.%04d.<frame>.<ext>`. I have not been able to get find the documentation about this but I am assuming the render settings for the about are baked in the `vrscene` file. There for I have removed the `%04d` from the output file name.

3. I added the a new collector which benefits both render and vray scene exporting. It will collect the renderable cameras for each renderlayer, this makes the validator for renderable cameras simpler.

### DIFF
--- a/colorbleed/plugins/maya/publish/submit_vray_deadline.py
+++ b/colorbleed/plugins/maya/publish/submit_vray_deadline.py
@@ -191,8 +191,11 @@ class VraySubmitDeadline(pyblish.api.InstancePlugin):
         cmd = ('-r vray -proj {project} -cam {cam} -noRender -s {startFrame} '
                '-e {endFrame} -rl {layer} -exportFramesSeparate')
 
+        # Get the camera
+        cammera = instance.data["cameras"][0]
+
         return cmd.format(project=instance.context.data["workspaceDir"],
-                          cam=instance.data.get("camera", "persp"),
+                          cam=cammera,
                           startFrame=instance.data["startFrame"],
                           endFrame=instance.data["endFrame"],
                           layer=instance.name)
--- a/colorbleed/plugins/maya/publish/submit_vray_deadline.py
+++ b/colorbleed/plugins/maya/publish/submit_vray_deadline.py
@@ -130,7 +130,7 @@ class VraySubmitDeadline(pyblish.api.InstancePlugin):
         instance.data["outputDir"] = render_ouput
 
         # Format output file name
-        sequence_filename = ".".join([instance.name, "%04d", ext])
+        sequence_filename = ".".join([instance.name, ext])
         output_filename = os.path.join(render_ouput, sequence_filename)
 
         payload_b = {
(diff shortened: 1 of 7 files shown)


colorbleed/plugins/maya/publish/submit_vray_deadline.py
```diff
@@ -191,15 +194,21 @@ def build_command(self, instance):
         cmd = ('-r vray -proj {project} -cam {cam} -noRender -s {startFrame} '
                '-e {endFrame} -rl {layer} -exportFramesSeparate')
 
+        # Get the camera
+        cammera = instance.data["cameras"][0]
```
BigRoy: Minor typo, double `mm` in `cammera`

colorbleed/plugins/maya/publish/validate_vray_translator_settings.py
```diff
@@ -13,6 +13,22 @@ class ValidateVRayTranslatorEnabled(pyblish.api.ContextPlugin):
 
     def process(self, context):
 
+        # Check if there are any vray scene instances
+        # The reason to not use host.lsattr() as used in collect_vray_scene
+        # is because that information is already available in the context
```
BigRoy: Should this comment also state why this workaround is implemented? That this code is there because of `pyblish-base#250`?

colorbleed/plugins/maya/publish/validate_render_single_camera.py
```diff
@@ -18,33 +15,19 @@ class ValidateRenderSingleCamera(pyblish.api.InstancePlugin):
...
-                               "found: {0}".format(invalid))
+
+    @classmethod
+    def get_invalid(cls, instance):
+        cameras = instance.data.get("cameras", [])
+        if len(cameras) != 1:
+            cls.log.error("Multiple renderable cameras" "found: %s " %
+                          instance.data["setMembers"])
+
+            return [instance.data["setMembers"]]
```
BigRoy: This should actually be returning the cameras. See the older logic from before.

colorbleed/plugins/maya/publish/validate_render_single_camera.py
```diff
@@ -18,33 +15,19 @@ class ValidateRenderSingleCamera(pyblish.api.InstancePlugin):
     """
 
     order = colorbleed.api.ValidateContentsOrder
+    label = "Render Single Camera"
     hosts = ['maya']
     families = ['colorbleed.renderlayer',
                 "colorbleed.vrayscene"]
-    label = "Render Single Camera"
-    actions = [colorbleed.maya.action.SelectInvalidAction]
```
BigRoy: Let's put this action back in so artists can automatically select which cameras are problematic. :)

PR #194, opened 2018-10-18
Added GenerateUUIDsOnInvalidAction to plugin

aardschok:
Resolves internal issue PLN-182

Added extra action to plugin to ensure user is able to create  valid IDs


BigRoy:
Looks good!

PR #196, opened 2018-10-18
Cleanup Creators, fix Yeti Cache preroll instance key and cosmetics

BigRoy:
- Cleanup Creators by removing redundant `OrderedDict()`
- Fix YetiCache instance attribute `preroll` (it had a typo and previously was `peroll`)
- Remove unused animation loader, delete `_load_animation.py`

---

@aardschok minor note, whenever we do:
```python
self.data.update({
  "a": 1,
  "b": 2,
  "c": 3
})
```

Then the data in the `self.data` (=OrderedDict) is not necessarily ordered a, b and then c. This is because we're updating with an intermediate regular `dict`: `{"a": 1, "b": 2, "c": 3}` which can have an arbitrary sorting prior to being added to the OrderedDict. :) As such it's important to understand that the order of the keys can be any order. That's why I refactored [this](https://github.com/Colorbleed/colorbleed-config/commit/9b7fc5302b0283508bf01f0cd1d0bd8a74a9d066) but there are some other areas where order remains "random" because of updating like that. Must we patch that up too?


aardschok:
It is a bit of a nit pick but it does look better if everything is in the same order in the instance as it is in code

PR #198, opened 2018-10-22
Fixed validator

aardschok:
_Resolves internal issue PLN-189_

Synchronized validator with the instance to ensure the validator does not fail on missing keys in the instance data.

For consistency I have added a `get_invalid` method similar to instance validators. The repair function can now be called without error. It needed to have a static method decorator instead of class method to ensure it could be triggered by the Repair Context Action.

All logging messages are now correctly called and visible in Pyblish.

### DIFF
--- a/colorbleed/plugins/maya/publish/validate_vray_translator_settings.py
+++ b/colorbleed/plugins/maya/publish/validate_vray_translator_settings.py
@@ -16,14 +16,8 @@ class ValidateVRayTranslatorEnabled(pyblish.api.ContextPlugin):
         # Check if there are any vray scene instances
         # The reason to not use host.lsattr() as used in collect_vray_scene
         # is because that information is already available in the context
-        vrayscene_instances = []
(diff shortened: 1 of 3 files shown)


colorbleed/plugins/maya/publish/validate_vray_translator_settings.py
```diff
@@ -36,17 +39,22 @@ def process(self, context):
...
-            self.info.error("Template for file name is wrong")
+            cls.log.error("Template for file name is wrong")
+            invalid = True
 
-    @classmethod
-    def repair(cls, context):
+        return invalid
+
+    @staticmethod
+    def repair():
```
BigRoy: Wait, if this works there must be something wrong. It should still receive the Context object as argument?

BigRoy:
@aardschok pushed a fix just now for the RepairContextAction - let me know if this is good. Feel free to merge.

PR #201, opened 2018-10-26
Collect machine name

aardschok:
Resolves internal issue PLN-170

Aside of Author or User include the Machine name that generated the publish. This helps debugging any specific machine issues, e.g. machine X created a wrong publish.

It will also currently help in showing who actually generated it since the user accounts on the workstations currently only list "Guest" or alike. :)


BigRoy:
Good stuff. :+1:

PR #203, opened 2018-10-26
Collect FPS from work scene

aardschok:
Resolves internal isse PLN-156

Publish along the FPS of the content (the units in the work scene) and make it visible in the loader.


aardschok:
See PR #204

PR #204, opened 2018-10-26
Add FPS check to Houdini pipeline

aardschok:
Adds FPS collectors for Houdini and Maya.

In order to have consistency across all DCCs I have added the `validate_fps` to the `before_save` callback of Houdini to ensure the project saved is the same setting as the project.


BigRoy:
Good! :+1: Aced it. Let's merge.

PR #210, opened 2018-11-12
PLN-198: Add quick icons for Loader/Manager in toolbox in Maya (draft)

BigRoy:


BigRoy:
maya_toolbox_icons

Above is a preview of the current state of the icons in the Toolbox.

PR #213, opened 2018-12-13
Exclude parent hierarchy by default for pointcaches

BigRoy:
By default exclude parent hierarchy for pointcaches (internal issue PLN-205)
This adds in a toggle to export alembics without the parent hierarchy

### DIFF
--- a/colorbleed/plugins/maya/create/colorbleed_model.py
+++ b/colorbleed/plugins/maya/create/colorbleed_model.py
@@ -1,5 +1,3 @@
-from collections import OrderedDict
-
 import avalon.maya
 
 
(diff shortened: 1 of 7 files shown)


colorbleed/plugins/maya/create/colorbleed_model.py
```diff
@@ -14,8 +12,12 @@ class CreateModel(avalon.maya.Creator):
     def __init__(self, *args, **kwargs):
         super(CreateModel, self).__init__(*args, **kwargs)
 
-        data = {"writeColorSets": False,  # Vertex colors with the geometry.
-                "attr": "",  # Add options for custom attributes
-                "attrPrefix": ""}
+        # Vertex colors with the geometry
```
BigRoy: This change for the existing is solely so that the order is preserved correctly for the attributes on the node. It's not required for this feature.

colorbleed/plugins/maya/publish/validate_model_content.py
```diff
@@ -63,7 +63,8 @@ def get_invalid(cls, instance):
             cls.log.error("Must have exactly one top group")
             if len(assemblies) == 0:
                 cls.log.warning("No top group found. "
-                                "(Are there objects in the instance?)")
+                                "(Are there objects in the instance?"
+                                " Or is it parented in another group?)")
```
BigRoy: This extra line in the warning message is to guide the artist more whenever the node is parented in a group that is not being published along.

PR #220, opened 2018-12-24
Implement maketx + hardlinking for publishing looks

BigRoy:
### Changes

This implements two major new functionalities for publishing looks, automatically generating `.tx` files and optimizing publishes by hardlinking when files are the same original file.

1. It will allow automatically generate `.tx` files on publish for optimized rendering. The files are generated with OpenImageIO's `maketx` [here](https://github.com/Colorbleed/colorbleed-config/compare/master...BigRoy:LKD-34?expand=1#diff-5a7f0286c4fd60a87c17fbaef37b6a67R58)
2. It will store a **sourceHash** per texture with the look publishes. The hash is generated [here](https://github.com/Colorbleed/colorbleed-config/compare/master...BigRoy:LKD-34?expand=1#diff-5a7f0286c4fd60a87c17fbaef37b6a67R22) and stored with publish [here](https://github.com/Colorbleed/colorbleed-config/compare/master...BigRoy:LKD-34?expand=1#diff-5a7f0286c4fd60a87c17fbaef37b6a67R272) and [here](https://github.com/Colorbleed/colorbleed-config/compare/master...BigRoy:LKD-34?expand=1#diff-6535682676747a834c2738e63b25a1d5R381)
3. Upon publishing whenever a matching **sourceHash** is found in the database it looks up the corresponding publish paths and whether they (still) exist (w/ `os.path.exists`), if so it will not reprocess for a `.tx` file and hardlink the previous publish.

- It is possible to disable the automatic `.tx` generation in the Look instance by setting the `maketx` attribute to False. The hashing will continue to work as it can differentiate between a previous source file being preprocessed by `maketx` or not, see [here](https://github.com/Colorbleed/colorbleed-config/compare/master...BigRoy:LKD-34?expand=1#diff-5a7f0286c4fd60a87c17fbaef37b6a67R310)

### The source hash

An example hash generated for a texture:
```
p:/projects/test1/assets/bank/work/lookdev/maya/images/art,png|1523361882,48|15302|maketx
```
- Dots are swapped with comma so it can be a valid pymongo key.
- `{path-to-source-lowercased-forwardslashed}|{modification time in seconds}|{size in bytes}|{*args}` (current hash breakdown)

_I didn't do any additional encoding to the hash (for now?) as I liked being able to read it for debugging purposes_.

---

#### What's next?

Some optimizations can still be made:
- Now it queries each texture on its own with the database, likely this can be optimized - nevertheless that would likely be a micro-optimization.
- If the hashing is changed in the future, previous hashes are not recognized anymore as they hash key lookup would turn out to be searching for a different value - but that should make sense.
- The "source file" hash currently stores the source original full path for more uniqueness, this could be brought down to just the source filename + modication date.


mkolar:
We've tested this on our end and it's working very well. The question obviously is what criteria to use for checking the textures. 

We found out that for our workflow it's often quite time consuming for artists to selectively update only certain textures in the maya workfile. Normally they re-export whole texture pack from substance and up the version in maya in bulk. making sure everything is new. Now because all the textures have been regenerated from substance, even the ones that visually didn't change will have slightly different size and would also end up with different hash. For small assets artist can keep track of this and only load the changed texture, however for the bigger assets this quickly becomes unmanagable for them, resorting full update again, hence losing the linking advantage.

However technically it works and we'll give it a shot in production.

BigRoy:
> We found out that for our workflow it's often quite time consuming for artists to selectively update only certain textures in the maya workfile. Normally they re-export whole texture pack from substance and up the version in maya in bulk. making sure everything is new. Now because all the textures have been regenerated from substance, even the ones that visually didn't change will have slightly different size and would also end up with different hash. For small assets artist can keep track of this and only load the changed texture, however for the bigger assets this quickly becomes unmanagable for them, resorting full update again, hence losing the linking advantage.

If you can get some control on Substance, then potentially before you "replace" the files you could do a hash comparison to see the files are actually the same. However, to do such a check you'd need to read the file and it'd be relatively slow (because of disk i/o). The same check could be done during publish, yet again... I didn't do that because it was way slower for most of our use cases.

colorbleed/plugins/maya/publish/extract_look.py
```diff
@@ -2,16 +2,103 @@
...
+
+    You can specify additional arguments in the function
+    to allow for specific 'processing' values to be included.
+
+    """
+    # We replace dots with comma because . cannot be a key in a pymongo dict.
+    return "|".join([
+                        os.path.normpath(os.path.abspath(filepath))
+                        .lower().replace("\\", "/"),
+                        str(os.path.getmtime(filepath)),
```
davidlatwe: I was snooping around, and although this is an old PR which might be out-dated, but since I got this problem a few days ago so I am just leaving a note here anyway :P

The file's modification time might lose a little accuracy after copied by Python, you may see this [post](https://stackoverflow.com/a/17087580/4145300) for detail.

So after a few tests with my site's storage server, I end up using this `math.floor(mtime * 10000) / float(10000)` to chop off the rest digit (floor at 4 decimal places) to get my *soft* mtime.

One other thing is you might want to use `shutil.copy2` to preserve the file's metadata (including mtime) when copying files.

Hopes this helps :D

BigRoy: Thanks for dropping those comments - definitely will help when/if we start adopting is. At this stage we never merged this for production as I didn't test it that thoroughly and the need at the time went away as a production got cancelled.

However, and thanks a ton for mentioning this one:

> The file's modification time might lose a little accuracy after copied by Python, you may see this post for detail.

That must be one that is insanely hard to track down if you wouldn't know exactly what was going on. So thanks, that will probably save me time. :D

mkolar: great comment. thanks. We've actually deployed this as it's causing no harm and saves some space. It doesn't quite catch all scenarios, but still makes a difference

BigRoy: Are you also using the `maketx` functionality? Or solely the hardlinking?

mkolar: maketx as well. we're only publishing .tx files. we also deal with colourspace transforms upon publishing. Taking them from the file node. Even though that part needs some love.

BigRoy: > we also deal with colourspace transforms upon publishing. Taking them from the file node. Even though that part needs some love.

Definitely point me to what you're addressing and how if you have the code available somewhere. :) But thanks for the details!

mkolar: The code is here, but it's very basic at this point. https://github.com/pypeclub/pype/blob/c20f4ba5ee7cc68022a9d0eaae2440a685d74ff3/pype/plugins/maya/publish/extract_look.py#L180

PR #221, opened 2018-12-31
Add Work Files app as icon in toolbar in Maya

BigRoy:


BigRoy:
This was actually related to internal issue PLN-198.

tokejepsen:
What icon did you use?

tokejepsen:
Nice and simple :-) Maybe we should use that in the core.

PR #249, opened 2019-08-20
Recover from #418

mottosso:
Repaired since having merged https://github.com/getavalon/core/pull/418. There's work left to do, see https://github.com/getavalon/core/issues/420


mottosso:
Oh, and the PyPI integration is in here too! :)

BigRoy:
Looks good to me, thanks. Will merge this now. Note that we are not using the master branch currently but the `acre` branch which has much more commits. Nevertheless I will pull this into our working state when I am back from vacation and was able to test things too.

PR #257, opened 2019-09-20
Implement customized animLayer baking for FBX family

BigRoy:
This implements an extra feature to the FBX family.

### Bake custom frame ranges to baked animLayers in the FBX

This will bake the members of the export to separate animation layers, for example:
```python
[
	["slide", 1, 25],
	["fly", 100, 120]
]
```

The above will bake out an animation layer `slide` from frame 1 through 25 and a layer `fly` from 100 through 120. Each layer will start at frame zero, the baked animation is shifted to always start at zero.

_Note: This code currently does not remove any already existing animation layers, so those might also end up in the FBX export separately_

Some more examples:

```python
# A single layer
[ 
    ["layername", 100, 200]
]

# Three animations, intro, middle and outro
[
    ["intro", 100, 200], 
    ["middle", 200, 300], 
    ["outro", 300, 400] 
]
```

These values can be edited on the FBX publish instance in the scene in the `bakeAnimLayers` attribute as a formatted `JSON` string.

I've also put a very basic example layout in an **online JSON editor [here](https://jsoneditoronline.org/?id=de5f018bb33c464aba321913d710501e)**

#### Local scene remains unaltered after export

After extraction the local scene is *unchanged* and the artist can continue to work as they were before.

### DIFF
--- a/colorbleed/houdini/lib.py
+++ b/colorbleed/houdini/lib.py
@@ -248,7 +248,7 @@ def create_remote_publish_node(force=True):
 
     """
 
-    cmd = "import pyblish.util; pyblish.util.publish()"
+    cmd = "import colorbleed.lib; colorbleed.lib.publish_remote()"
 
     existing = hou.node("/out/REMOTE_PUBLISH")
     if existing:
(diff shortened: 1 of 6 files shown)


BigRoy:
Pushed another feature to the same branch just now, bit messy. Sorry.

Anyway. I've moved the remote publish logic to `colorbleed.lib` so that now Houdini and Maya perform the same remote publish logic.

